PR #10, opened 2024-09-06
bug[TW27114]: PUB - Fix view not passed from client to server publish

jadenpuckett:
Change-Id: Ib7b4747b7629d8def2ad885c6c1af222fbc6713d

### DIFF
--- a/plugins/org.eclipse.osee.framework.ui.skynet/src/org/eclipse/osee/framework/ui/skynet/render/FileSystemRenderer.java
+++ b/plugins/org.eclipse.osee.framework.ui.skynet/src/org/eclipse/osee/framework/ui/skynet/render/FileSystemRenderer.java
@@ -28,7 +28,6 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.jface.resource.ImageDescriptor;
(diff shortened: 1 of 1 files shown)


audrey-denk (approved):

plugins/org.eclipse.osee.framework.ui.skynet/src/org/eclipse/osee/framework/ui/skynet/render/FileSystemRenderer.java
```diff
@@ -28,7 +28,6 @@
 import org.eclipse.jdt.annotation.NonNull;
```
NerolYelhsa, line 28: I am worried that this might loose the view for other entry paths through the RendererManager. If every thing checks out this is a good simplification. I'm not sure that the regression test suite is sufficient to catch it though. Perhaps we should still set the view if it has not already been set in the renderer options.

jadenpuckett: The view is set through another UI element and its value is stored in the initial creation of renderer options elsewhere. This method's purpose is to ensure all artifacts are from the same branch. The loaded branchToken is not expected to have a view inside, hence the already set view in the renderer options is being set to sentinel (-1) every time.

dondunne (approved):

PR #23, opened 2024-09-16
bug[TW27160]: Fix TW being created without name

audrey-denk:
Change-Id: Ia3a78ff5890a251608c57193857b115512817a31


ryanbase (approved):

lucianovaglienti (approved):

plugins/org.eclipse.osee.ats.core/src/org/eclipse/osee/ats/core/action/AtsActionService.java
```diff
@@ -545,9 +544,10 @@ public IAtsTeamWorkflow createTeamWorkflow(IAtsAction action, IAtsTeamDefinition
             artToken = listener.getArtifactToken(applicableAis);
          }
       }
-
+      
       if (artToken == null) {
-         teamWf = atsApi.getWorkItemService().getTeamWf(changes.createArtifact(artifactType, ""));
+         String name = (action.getName() == null) ? Strings.EMPTY_STRING : action.getName();
```
lucianovaglienti, line 549: Question...is there a sensible default we could use instead? Like TW(NEXT_NUMBER_HERE)?

PR #55, opened 2024-10-10
bug[TW27333]: CreateNewChangeRequestBlam executes if errors

dondunne:


plugins/org.eclipse.osee.ats.ide/src/org/eclipse/osee/ats/ide/actions/OpenOrphanedActionsAndTeamWorkflows.java
```diff
@@ -0,0 +1,93 @@
+/*********************************************************************
+ * Copyright (c) 2016 Boeing
```
Grumb1e, line 2: Old distribution statement year.

Grumb1e (commented):
Do we have a test that creates a CR and verifies that is created correctly? I want to ensure similar CR issues are prevented in the future.

plugins/org.eclipse.osee.ats.ide/src/org/eclipse/osee/ats/ide/actions/OpenOrphanedActionsAndTeamWorkflows.java
dondunne: Yes, but the offending field wasn't added to the test correctly, so it didn't fail.  That field has been removed from the creation, so it's OBE.

Grumb1e (approved):

PR #65, opened 2024-10-17
bug[TW26612]: Last Modified date not showing for relations

dondunne:
Change-Id: Icee39090acecb01486b3b79eab8d0fd0503b668b


plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/artifact/RelationLoader.java
```diff
@@ -93,7 +108,20 @@ public static void loadRelationData(Long joinQueryId, Collection<Artifact> artif
...
 
       for (Artifact artifact : artifacts) {
+         setTransactionRecord(artToTx, artifact);
          artifact.setLinksLoaded(true);
       }
    }
+
+   private static void setTransactionRecord(Map<ArtifactToken, TransactionToken> artToTx, Artifact artifact) {
+      TransactionToken txTok = artToTx.get(artifact);
+      if (txTok != null && txTok.isValid()) {
```
lucianovaglienti, line 118: This should never be null if it is coming from the DB since me and Audrey made the change to prevent strings being nulled improperly.

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/artifact/RelationLoader.java
```diff
@@ -93,7 +108,20 @@ public static void loadRelationData(Long joinQueryId, Collection<Artifact> artif
...
+         TransactionRecord txRec = TransactionManager.getTransactionRecord(txTok);
```
lucianovaglienti, line 119: Why is this needed? You've already got the TransactionToken above from the map(which is loaded from the db)

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/artifact/RelationLoader.java
```diff
@@ -93,7 +108,20 @@ public static void loadRelationData(Long joinQueryId, Collection<Artifact> artif
...
+         if (txRec != null && txRec.isValid()) {
```
lucianovaglienti, line 120: Same as above, null case probably doesn't exist.

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/transaction/TransactionManager.java
```diff
@@ -471,4 +475,19 @@ public static void persistInTransaction(String comment, Artifact... artifacts) {
       transaction.execute();
    }
 
+   public static TransactionRecord getTransactionRecord(TransactionId txId) {
```
lucianovaglienti, line 478: Provided the RelationLoader comments prove to be true and you don't end up needing this there...Do we need this? It would only be used in a test at that point. + it's always good to get rid of caches :)

lucianovaglienti (commented):
Some minor comments...

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/artifact/RelationLoader.java
dondunne: This is a cache so don't have to load the TransactionRecord multiple times per artifact.

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/artifact/RelationLoader.java
dondunne: TransactionRecord has all the data about transaction and is on every client Artifact.  Cheep to get and cache as transactions will never change and it keeps UI from having to load in all the places it needs the record.

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/transaction/TransactionManager.java
dondunne: TransactionRecords don't ever change, so this cache isn't bad.  Agreed on getting rid of caches that have to be kept up to date or cleared.

jadenpuckett (approved):

smolaro (approved):

PR #66, opened 2024-10-17
feature[TW27211]: Add create version endpoint

audrey-denk:
Change-Id: Ib7b04326b3aee6ad52eb81b81f56b45dd74b3e92

### DIFF
--- a/plugins/org.eclipse.osee.ats.api/src/org/eclipse/osee/ats/api/config/AtsConfigEndpointApi.java
+++ b/plugins/org.eclipse.osee.ats.api/src/org/eclipse/osee/ats/api/config/AtsConfigEndpointApi.java
@@ -22,6 +22,7 @@ import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 import org.eclipse.osee.ats.api.ai.ActionableItem;
 import org.eclipse.osee.ats.api.branch.BranchData;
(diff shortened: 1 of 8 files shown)


plugins/org.eclipse.osee.ats.api/src/org/eclipse/osee/ats/api/config/AtsConfigEndpointApi.java
```diff
@@ -137,6 +138,12 @@ public interface AtsConfigEndpointApi {
    @Produces(MediaType.APPLICATION_JSON)
    public Version getVersion(@PathParam("verId") ArtifactId verId);
 
+   @POST
+   @SkipAtsConfigJsonWriter
```
dondunne, line 142: Remove the "/create".  That is assumed by POST.  The method can still be createVersion.

plugins/org.eclipse.osee.ats.core/src/org/eclipse/osee/ats/core/version/AtsVersionServiceImpl.java
```diff
@@ -327,6 +327,16 @@ public Collection<IAtsVersion> getVersions(IAtsTeamDefinition teamDef) {
    public Version createVersion(String title, IAtsChangeSet changes) {
       return createVersion(title, Lib.generateArtifactIdAsInt(), changes);
    }
+   
+   @Override
+   public Version createVersion(String name, String description, ArtifactId teamId) {
+      IAtsChangeSet changes = atsApi.createChangeSet("Create Version: "+name);
+      Version newVersion = createVersion(name,changes);
+      newVersion.setDescription(description);
```
dondunne, line 335: I'd load the teamId and make sure it exists and isOfType AtsArtifactTypes.TeamDefinition.  Exception if it doesn't.

david-w-miller (commented):
Quick checks - since anything could be coming in, can we check to make sure the name and description are reasonable?

plugins/org.eclipse.osee.ats.api/src/org/eclipse/osee/ats/api/config/AtsConfigEndpointApi.java
audrey-denk: I thought of that but in some cases we are explicit.  I'll update it.

dondunne (approved):

david-w-miller (approved):

PR #80, opened 2024-10-24
bug[TW27438]: Resolve relations npe when artifact explorer web querie…

jadenpuckett:
…s for artifact with relations


Grumb1e (approved):

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/data/ArtifactWithRelations.java
```diff
@@ -28,18 +28,23 @@ public class ArtifactWithRelations {
    private final ArtifactReadable artReadable;
    private final OrcsTokenService tokenService;
    private final List<ArtifactWithRelationsAttribute> attributes = new ArrayList<>();
-   private final boolean includeRelations;
+   private List<ArtifactRelation> relations;
```
audrey-denk, line 31: Maybe initialize here?  = Collections.EmptyList();

PR #93, opened 2024-11-05
Upgrade Eclipse Version To 2022-06

Grumb1e:


jadenpuckett (approved):

smolaro (approved):

lucianovaglienti (approved):
+1 once you can confirm ng e2e validates the server :)

Grumb1e:
Ran the e2e tests. Some of those tests are still under development and others need to be made more variance tolerant. There is some indication of expected operation for the server though. Additional changes will almost certainly be required, but this would suggest things are functioning at a base level.

PR #124, opened 2024-11-25
Replace getRelationTypes with getValidRelationTypes…

audrey-denk:
… to prevent dups from showing

Change-Id: I0cac874d9ed1c25eaf13fd611afaf2cf8020d583


lucianovaglienti (approved):

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/data/OrcsTokenServiceImpl.java
```diff
@@ -437,6 +437,40 @@ public Set<ArtifactTypeToken> getValidArtifactTypes(AttributeTypeToken attribute
       return artifactTypes;
    }
 
+   @Override
+   public List<RelationTypeToken> getValidRelationTypes() {
+      Collection<RelationTypeToken> relationTypes = getRelationTypes();
+      List<RelationTypeToken> invalidRels = new ArrayList<>();
+      //populate convertedRelations if empty
+      if (convertedRelations.isEmpty()) {
```
lucianovaglienti, line 445: Shouldn't this be done elsewhere?

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/data/OrcsTokenServiceImpl.java
```diff
@@ -437,6 +437,40 @@ public Set<ArtifactTypeToken> getValidArtifactTypes(AttributeTypeToken attribute
...
+         setConvertedRelations();
+      }
+      //add all rels in shadowRels and remove the ones that have been converted
+      for (Field field : Arrays.asList(ShadowCoreRelationTypes.class.getDeclaredFields())) {
+         try {
+            RelationTypeToken rel = (RelationTypeToken) field.get(null);
```
lucianovaglienti, line 451: What is this doing?

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/data/OrcsTokenServiceImpl.java
audrey-denk: Its done the first time this is called OR the first time a specific relation is called out.

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/data/OrcsTokenServiceImpl.java
audrey-denk: This is for when we start actually converting relations.  Once DefaultHierarchy is converted for example, we would remove the "old" one from what can be returned as a valid relation.

ryanbase (approved):

PR #125, opened 2024-11-25
feature[TW27628]: Fix Recently Visited Items in Navigator

dondunne:
Change-Id: I95a5f2f40040d91fb53ddccb752dcee955160482


plugins/org.eclipse.osee.ats.api/src/org/eclipse/osee/ats/api/util/RecentlyVisistedItem.java
```diff
@@ -0,0 +1,101 @@
+/*********************************************************************
+ * Copyright (c) 2019 Boeing
```
vaibhavypatel, line 2: Change it to 2024 maybe ?

vaibhavypatel (approved):

PR #128, opened 2024-11-26
feature[TW27637]: Fix missing item indexer to include single char values

audrey-denk:
Change-Id: I0e36047277a69a3f188e171476a15fa820a167f5


dondunne (approved):

david-w-miller (approved):

david-w-miller:
Done - I didn't check to see if way in the past it was 0, then someone did some fix to improve performance and put it to 1, thinking that we shouldn't be searching for one character, improving the performance by eliminating a whole bunch of attributes, but breaking all of those for search.

PR #134, opened 2024-12-04
bug[TW27664]: Branch Type buttons missing from Branch Search

dondunne:
Change-Id: I4bc2619febae60c26278f81df6735505e58d327a


plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/User.java
```diff
@@ -127,18 +133,23 @@ public void toggleFavoriteBranch(BranchId favoriteBranch) {
...
-      for (String value : attributes) {
-         try {
-            if (branch.equals(BranchId.valueOf(value))) {
-               return true;
-            }
-         } catch (Exception ex) {
-            // do nothing
+      if (!getTransaction().equals(userModTx)) {
+         if (favoriteBranchIds == null) {
+            favoriteBranchIds = new HashSet<>(20);
```
jadenpuckett, line 143: is there a size limit of 20? why is this 20?

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/User.java
```diff
@@ -127,18 +133,23 @@ public void toggleFavoriteBranch(BranchId favoriteBranch) {
...
+         } else {
+            favoriteBranchIds.clear();
+         }
+         for (Attribute<Object> attri : getAttributes(CoreAttributeTypes.FavoriteBranch)) {
```
jadenpuckett, line 147: is it worth taking the time to make the FavoriteBranch attribute an AttributeTypeLong and not an AttributeTypeString to avoid converting every time we store/fetch?

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/User.java
dondunne: There to give default size of HashSet if you know it's going to be a certain size so Set doesn't have to be recreated as it grows.  Probably not necessary in this case since most don't have favorites and those that do probably not many.  Removed "20".

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/User.java
dondunne: Yes, love the idea.  Want to get this in to resolve the immediate lockups and missing buttons.  Will make an action to convert which will take a little time and some validation of data in dbs.

jadenpuckett: Sounds great!

PR #141, opened 2024-12-11
bug[TW27720]: PUB - Update publisher to return empty document with me…

jadenpuckett:
…ssage when artifact(s) is not applicable for view

### DIFF
--- a/plugins/org.eclipse.osee.define/src/org/eclipse/osee/define/operations/publisher/publishing/GeneralPublishingWordTemplateProcessorServer.java
+++ b/plugins/org.eclipse.osee.define/src/org/eclipse/osee/define/operations/publisher/publishing/GeneralPublishingWordTemplateProcessorServer.java
@@ -71,98 +71,105 @@ public class GeneralPublishingWordTemplateProcessorServer extends WordTemplatePr
             PublishingAppender       wordMl
          ) {
 
-      this.emptyFoldersArtifactAcceptor =
-         WordRenderUtil.populateEmptyHeaders
-            (
-               artifacts,
-               this.includeHeadings,
-               this.allowedOutlineTypes,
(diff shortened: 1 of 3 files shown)


Grumb1e (dismissed):

Grumb1e (approved):

plugins/org.eclipse.osee.define/src/org/eclipse/osee/define/operations/publisher/publishing/GeneralPublishingWordTemplateProcessorServer.java
```diff
@@ -71,98 +71,105 @@ public GeneralPublishingWordTemplateProcessorServer(OrcsApi orcsApi, AtsApi atsA
...
+          * <p>
+          * Given the list of artifacts for the publish, this loops through and adds any recursive artifacts to also be
+          * published (if specified through recurseChildren) and determines all of their data rights.
+          */
+
+         //@formatter:off
+         WordRenderUtil
+            .getDataRights
+               (
+                  /*
```
ryanbase, line 107: If we're going to have this much javadoc for the parameters, it should be on the method declaration as @param rather than each individual param when calling it.

ryanbase (approved):

plugins/org.eclipse.osee.define/src/org/eclipse/osee/define/operations/publisher/publishing/GeneralPublishingWordTemplateProcessorServer.java
jadenpuckett: They are also defined as @params in the .getDataRights javadoc... but I am hesitant to remove any javadoc that I didn't write just yet

Grumb1e (approved):

PR #162, opened 2025-01-08
feature[TW27811]: Update branching option to use txCurrent in addition

audrey-denk:
Using transaction_id ordering method takes up to 5 minutes to create a branch.. TxCurrent has been proven to be up to date and performance in seconds; therefore should be used instead.

Verified that Transactions are atomic, so if a working branch is being created at the same time as another is being merged, there should be no partial states.

### DIFF
--- a/plugins/org.eclipse.osee.ats.ide.integration.tests/src/org/eclipse/osee/ats/ide/integration/tests/skynet/core/BranchStateTest.java
+++ b/plugins/org.eclipse.osee.ats.ide.integration.tests/src/org/eclipse/osee/ats/ide/integration/tests/skynet/core/BranchStateTest.java
@@ -235,7 +235,13 @@ public class BranchStateTest {
       BranchToken workingBranch = null;
       try {
          workingBranch = BranchManager.createWorkingBranch(SAW_Bld_1, originalBranchName);
+         Artifact change = ArtifactTypeManager.addArtifact(CoreArtifactTypes.SoftwareRequirementMsWord, workingBranch,
+            "A commit change");
+         change.persist(getClass().getSimpleName());
 
+         Artifact workingBranchRoot = OseeSystemArtifacts.getDefaultHierarchyRootArtifact(workingBranch);
+         workingBranchRoot.addChild(change);
+         workingBranchRoot.persist(getClass().getSimpleName());
          // Update the branch
          ConflictResolverOperation resolverOperation =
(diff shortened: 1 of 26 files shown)


lucianovaglienti (dismissed):

dondunne (approved):

plugins/org.eclipse.osee.ats.ide.integration.tests/src/org/eclipse/osee/ats/ide/integration/tests/skynet/core/BranchStateTest.java
```diff
@@ -235,7 +235,13 @@ public void testRebaselineBranchNoChanges() throws Exception {
       BranchToken workingBranch = null;
       try {
          workingBranch = BranchManager.createWorkingBranch(SAW_Bld_1, originalBranchName);
+         Artifact change = ArtifactTypeManager.addArtifact(CoreArtifactTypes.SoftwareRequirementMsWord, workingBranch,
+            "A commit change");
+         change.persist(getClass().getSimpleName());
 
+         Artifact workingBranchRoot = OseeSystemArtifacts.getDefaultHierarchyRootArtifact(workingBranch);
+         workingBranchRoot.addChild(change);
+         workingBranchRoot.persist(getClass().getSimpleName());
```
lucianovaglienti, line 244: Just kinda wondering...what's the intention of this test? If it's testing whether or not you can rebaseline a branch that has no changes, wouldn't you modify the underlying baseline branch then trigger a rebaseline?

lucianovaglienti (approved):

plugins/org.eclipse.osee.ats.ide.integration.tests/src/org/eclipse/osee/ats/ide/integration/tests/skynet/core/BranchStateTest.java
audrey-denk: Yeah.. this one threw me for a loop.  It isn't documented well and maybe I should add documentation.  I think it is testing updating from parent when there are no updates to the parent but there are updates to the working branch.  Before my change it was relying on the branch category change somehow... anyhow, in my quest to figure it out I realized I needed to update the query to get the gammas to exclude the branch category

ryanbase (approved):

PR #166, opened 2025-01-10
feature[TW27840]: Disable creating backup files for development

dondunne:
Change-Id: Iaf1bce55abdb06ebf33bb1081e26f7193e6759e2

### DIFF
--- a/plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionFactoryImpl.java
+++ b/plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionFactoryImpl.java
@@ -195,8 +195,8 @@ public class TransactionFactoryImpl implements TransactionFactory {
       for (TransactionId txIdToDelete : txsToDelete) {
          BranchId branchId = orcsApi.getTransactionFactory().getTx(txIdToDelete).getBranch();
          orcsApi.getJdbcService().getClient().runQuery(stmt -> insertTxStatements.add(stmt.getString("insertString")),
-            OseeDb.TX_DETAILS_TABLE.getSelectInsertString(" where branch_id = ? and transaction_id = ?"),
-            branchId, txIdToDelete);
+            OseeDb.TX_DETAILS_TABLE.getSelectInsertString(" where branch_id = ? and transaction_id = ?"), branchId,
+            txIdToDelete);
          orcsApi.getJdbcService().getClient().runQuery(stmt -> insertTxStatements.add(stmt.getString("insertString")),
             OseeDb.TXS_TABLE.getSelectInsertString(" where branch_id = ? and transaction_id = ?"), branchId,
             txIdToDelete);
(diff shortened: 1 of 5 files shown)


plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionFactoryImpl.java
```diff
@@ -195,8 +195,8 @@ public boolean purgeTxs(String txIds) {
       for (TransactionId txIdToDelete : txsToDelete) {
          BranchId branchId = orcsApi.getTransactionFactory().getTx(txIdToDelete).getBranch();
          orcsApi.getJdbcService().getClient().runQuery(stmt -> insertTxStatements.add(stmt.getString("insertString")),
-            OseeDb.TX_DETAILS_TABLE.getSelectInsertString(" where branch_id = ? and transaction_id = ?"),
-            branchId, txIdToDelete);
+            OseeDb.TX_DETAILS_TABLE.getSelectInsertString(" where branch_id = ? and transaction_id = ?"), branchId,
```
audrey-denk, line 198: I would think you may want to skip all the queries that get impactedGammas altogether if in development? Although, I am a bit worried that we won't be able to test that the zip files are being created?

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionFactoryImpl.java
```diff
@@ -195,8 +195,8 @@ public boolean purgeTxs(String txIds) {
...
+            txIdToDelete);
```
audrey-denk, line 199: Does this need to change since you are checking for isDbProduction in the places where the actualy storage of the backup files takes place?

plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/health/PurgeUnusedBackingDataAndTransactions.java
```diff
@@ -215,34 +214,36 @@ public int[] purgeUnused(int rowCount) {
...
-            PrintWriter textWriter = new PrintWriter(writer);
-            addFileToZip(zipOut, OseeDb.ARTIFACT_TABLE.getName(), insertStatements, textWriter);
-            addFileToZip(zipOut, OseeDb.ATTRIBUTE_TABLE.getName(), insertStatements, textWriter);
-            addFileToZip(zipOut, OseeDb.RELATION_TABLE.getName(), insertStatements, textWriter);
-            addFileToZip(zipOut, OseeDb.RELATION_TABLE2.getName(), insertStatements, textWriter);
-            addFileToZip(zipOut, OseeDb.OSEE_SEARCH_TAGS_TABLE.getName(), insertStatements, textWriter);
-            zipOut.putNextEntry(new ZipEntry("ServerResourcesToDelete.txt"));
-            for (String string : uriResourcesToDelete) {
-               textWriter.println(string);
+      if (jdbcClient.getConfig().isProduction()) {
```
audrey-denk, line 217: This seems fine; however I wonder if we added tests to see if backups are happening, this would not allow them to work?  Not sure what the solution is, cause I understand wanting to clean up.  I wonder if perhaps we create something that runs after TestSuites to clean up?

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionFactoryImpl.java
dondunne: I check for production here cause the purge operation doesn't have orcsApi.  So if prefix is empty string, then the backups aren't done.

plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/health/PurgeUnusedBackingDataAndTransactions.java
dondunne: I was trying to save time during tests and runtime by not doing the operations at all.  If we add tests, we'll have to re-think this.

vaibhavypatel (approved):

audrey-denk (approved):

PR #182, opened 2025-01-30
feature[TW27822]: SCA - Identify and Display Code File Path

smolaro:
Change-Id: I8a22914b19331936f01bebf70cf0ae984af65e24

### DIFF
--- a/plugins/org.eclipse.osee.disposition.rest/src/org/eclipse/osee/disposition/rest/internal/OrcsStorageImpl.java
+++ b/plugins/org.eclipse.osee.disposition.rest/src/org/eclipse/osee/disposition/rest/internal/OrcsStorageImpl.java
@@ -203,10 +203,7 @@ public class OrcsStorageImpl implements Storage {
    }
 
    private ArtifactReadable findDispoArtifact(BranchId branch, String artId) {
-      return getQuery()//
-         .fromBranch(branch)//
-         .andUuid(Long.valueOf(artId))//
-         .getResults().getExactlyOne();
+      return getQuery().fromBranch(branch).andId(ArtifactId.valueOf(artId)).getResults().getExactlyOne();
    }
 
    @Override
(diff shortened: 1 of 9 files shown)


plugins/org.eclipse.osee.disposition.rest/src/org/eclipse/osee/disposition/rest/internal/OrcsStorageImpl.java
```diff
@@ -203,10 +203,7 @@ public DispoSet findDispoSetsById(BranchId branch, String setId) {
    }
 
    private ArtifactReadable findDispoArtifact(BranchId branch, String artId) {
-      return getQuery()//
-         .fromBranch(branch)//
-         .andUuid(Long.valueOf(artId))//
-         .getResults().getExactlyOne();
+      return getQuery().fromBranch(branch).andId(ArtifactId.valueOf(artId)).getResults().getExactlyOne();
```
justinjschaub, line 206: Can you describe the practical impact of change from `andUuid()` to `andId`? Within the data model, the ID and the UUID are not likely interchangeable are they? Is this change fixing a bug?

smolaro: Its effectively the same. `andUuid()` call requires us to convert from string to long to use. It then converts that long to an ArtifactId.
I wanted to make it consistent with the majority of the code base.

jadenpuckett (approved):

plugins/org.eclipse.osee.disposition.rest/src/org/eclipse/osee/disposition/rest/internal/OrcsStorageImpl.java
dondunne: yes, we started with string ids (guids), then converted to integers/longs (uuid and id).  We've been gradually converting all integers to longs and using the term id to reduce confusion.

dondunne (approved):

PR #190, opened 2025-02-05
bug[TW27528]: Fix Web URL for referencing artifacts

vaibhavypatel:


plugins/org.eclipse.osee.framework.ui.skynet/src/org/eclipse/osee/framework/ui/skynet/artifact/editor/action/CopyArtifactURLAction.java
```diff
@@ -77,8 +77,8 @@ public static boolean isApplicable(Artifact artifact) {
 
    public String generateLink() {
       return String.format("%sorcs/branch/%s/artifact/%s/attribute/type/%s",
```
Grumb1e, line 79: Will a / be needed before orcs here when using the osee.web.url? Also when defaulting to the prior, I assume it returns with a / already.

Grumb1e: Something along these lines might be able to handle either case 

String baseURL = System.getProperty("osee.web.url", new ArtifactUrlClient().getSelectedPermanentLinkUrl());
URI baseUri = new URI(baseURL);
URI resolvedUri = baseUri.resolve(path);

vaibhavypatel: Good Catch Resolved

Grumb1e (approved):

Grumb1e (approved):

dondunne (approved):

PR #191, opened 2025-02-06
feature[TW27878]: Fix Spotbugs Errors

torin-grenda:


plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/publishing/markdown/MarkdownPublishingAppender.java
```diff
@@ -295,10 +295,9 @@ public void startListPresentation() {
          ;
       //@formatter:on
    }
-   
+
    @Override
    public void endOutlineSubSection() {
-       this.endOutlineSubSection();
```
Grumb1e, line 301: I can see the trouble in what it was, but does it make sense to change the function to do nothing?

Grumb1e: Might be worth showing this change to someone in publishing like @jadenpuckett

jadenpuckett: Just checked. Its un-implemented like many markdown appender methods. Good to remove.

jadenpuckett (approved):

Grumb1e (approved):

PR #201, opened 2025-02-19
feature[TW28022]:  Fix sheet naming issues and struct/submsg rels

audrey-denk:
Update to validate struct/submsg relation and add hash to end of duplicate sheet name.


plugins/org.eclipse.osee.mim/src/org/eclipse/osee/mim/internal/MimIcdGenerator.java
```diff
@@ -1487,8 +1489,8 @@ private int[] printHeaderStructureElementRow(ExcelWorkbookWriter writer, AtomicI
       Integer endByte = Math.floorMod(byteLocation + byteSize - 1, 4);
       String elementName = element.getName().getValue();
       String dataType = logicalTypeToDataType(element.getPlatformType().getInterfaceLogicalType().getValue());
-      String units =
```
lucianovaglienti, line 1490: This formatting change intentional?

lucianovaglienti (approved):

plugins/org.eclipse.osee.mim/src/org/eclipse/osee/mim/internal/MimIcdGenerator.java
audrey-denk: No, but its fine... the line was way too long

lucianovaglienti (approved):

david-w-miller (approved):

PR #215, opened 2025-03-05
feature[TW27820]: Update OrcsWriter to overwrite default attribute va…

jadenpuckett:
…lue for enumerated attribute(s)

### DIFF
--- a/plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writer/OrcsCollectorWriter.java
+++ b/plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writer/OrcsCollectorWriter.java
@@ -17,6 +17,7 @@ import java.util.Calendar;
 import java.util.Collection;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


ryanbase (dismissed):

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writer/OrcsCollectorWriter.java
```diff
@@ -384,44 +385,58 @@ private void createAttributes(OwArtifact owArtifact, ArtifactToken artifact, XRe
             AttributeTypeToken attrType = getAttributeType(orcsApi.tokenService(), owAttrType);
 
             List<Object> values = owAttribute.getValues();
-            for (Object value : values) {
-               String valueOf = String.valueOf(value);
-               if (Strings.isValid(valueOf) && !valueOf.equals("null")) {
-                  if (attrType.isDouble()) {
```
dondunne, line 390: setSoleAttributeValue is going to overwrite all values so you'll only get the last one?

jadenpuckett: It will do an overwrite for the first value in the value list (the list of values for the current attribute type), then it will create attributes for subsequent values. For example, if value1 and value2 are in the imported spreadsheet, we want value1 to overwrite 'Unspecified' (or whatever the default value is), then we want value2 (and subsequent values) to use createAttribute to add another attribute with value2 and so on.

jadenpuckett:
Need to add test to cover the case in which enumerated attributes are imported using OrcsWriter

dondunne (approved):

ryanbase (approved):

PR #220, opened 2025-03-07
bug[TW27862]: Test adding SpotBugs to CI

torin-grenda:


CompSciDude1:
I want SpotBugs support in OTE to do static analysis on the test scripts in order to mitigate nasty logical errors that end up breaking the tests.

PR #223, opened 2025-03-11
feature[TW28054]: Improvements to XResultData and AtsChangeSet

dondunne:
Change-Id: I626a52975f4a30cfeb630d193cb34c2172c538e5


jadenpuckett (approved):

Grumb1e (commented):
This should be reviewed by people familiar with the web because I believe this is a type that is used there. It can be sensitive to changes and may break things.

Grumb1e (approved):
Talked with Luciano and Jaden. They believe it is likely fine because it is only additive. 

Should be checked to sure though.

PR #245, opened 2025-04-03
feature[TW28319]: Add Applicability to Artifact Editor

jadenpuckett:


smolaro (dismissed):

web/apps/osee/src/app/ple/artifact-explorer/lib/components/editor/attributes-editor-panel/attributes-editor-panel.component.html
```diff
@@ -33,13 +36,21 @@
 					>
 				</button>
 			</div>
-
+			<!--Applicability Dropdown-->
+			<div class="tw-flex tw-items-center tw-justify-between tw-pb-4">
+				<div class="tw-w-1/6 tw-pl-4 tw-font-bold">Applicability:</div>
+				<osee-persisted-applicability-dropdown
```
lucianovaglienti, line 42: FYI: you need a branch override for this component because these all assume ui.BranchId is the current branch

web/apps/osee/src/app/applicability/components/persisted-applicability-dropdown/persisted-applicability-dropdown.component.ts
```diff
@@ -93,7 +95,8 @@ export class PersistedApplicabilityDropdownComponent {
 						comment,
 						artId,
 						applic,
-						{}
+						{},
+						this.branchIdOverride()
```
lucianovaglienti, line 99: This should probably be handled as an input + computed i.e. 
```ts
branchIdOverride = input<string>(branchSentinel.id);
protected branchId = toSignal(this.uiService.id); 
protected selectedBranch = computed(()=> {
	const override = this branchIdOverride();
	if Number(override) > 0 {
		return override
	}
	return this.branchId
});

```

web/apps/osee/src/app/transactions/services/current-transaction.service.ts
```diff
@@ -177,11 +187,16 @@ export class CurrentTransactionService {
...
 	) => {
-		return this.modifyArt(comment, artId, applicability, attrConfig).pipe(
-			this.performMutation()
-		);
+		return this.modifyArt(
+			comment,
+			artId,
+			applicability,
+			attrConfig,
+			branchIdOverride
```
lucianovaglienti, line 198: Not really a fan of having an "override" in the transaction logic. I think a better approach would be to have an overload of function signatures and in the "base" function implementation, check if it is set to determine the right id to use.

jadenpuckett: Am i not doing an overload by adding a optional parameter to the methods?

lucianovaglienti: Functionally, yes, semantically no. An overload would be like fn function(params1); fn function(params2); fn function(params1orparams2){ //impl}

audrey-denk (approved):

web/apps/osee/src/app/transactions/services/current-transaction.service.ts
jadenpuckett: Ok so the impl will still have the optional branchIdOverride. Still just 1 impl. I will do this for the 3 methods that I have to pass the branchIdOverride through.

jadenpuckett: All I will do is implement 2 separate definitions for each method. Still 1 impl for each method.

Grumb1e (approved):

PR #271, opened 2025-04-30
feature[TW26905]: Implement image artifact renderer for Markdown publish

Grumb1e:


plugins/org.eclipse.osee.define/src/org/eclipse/osee/define/operations/markdown/MarkdownHtmlUtil.java
```diff
@@ -0,0 +1,129 @@
...
+         }
+         zipInputStream.closeEntry();
+      }
+
+      return new HtmlZip(imageContentMap, htmlDocument);
+   }
+
+   public static Node readMarkdown(InputStream inputStream) {
+      MutableDataSet options = new MutableDataSet();
+      options.set(Parser.EXTENSIONS, Arrays.asList(TablesExtension.create(), TaskListExtension.create(),
```
jadenpuckett, line 79: I would prefer if you set/read these options from a single common place in the code because we are reusing this line in 3+ places now. Maybe a Markdown.java? Flexmark.java? FlexmarkOptions.java? idk

dondunne (approved):

smolaro (approved):

plugins/org.eclipse.osee.define/src/org/eclipse/osee/define/operations/markdown/MarkdownHtmlUtil.java
Grumb1e: Will do! I will push up a small refactor commit. 👍

PR #274, opened 2025-05-01
Jaden/create branch

jadenpuckett:


plugins/org.eclipse.osee.ajax.deps/.project
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lucianovaglienti, line 1: intentional?

plugins/org.eclipse.osee.define/src/org/eclipse/osee/define/operations/publisher/publishing/WordRenderArtifactWrapperServerImpl.java
```diff
@@ -110,8 +111,8 @@ public WordRenderArtifactWrapperServerImpl(Object artifact) {
 
       var artifactBranchToken = this.artifact.getBranch();
 
-      this.branchToken =
```
lucianovaglienti, line 113: Can't we just update the existing BranchToken.create() that is called here to create the emptyList?

plugins/org.eclipse.osee.define/src/org/eclipse/osee/define/operations/publisher/publishing/WordRenderArtifactWrapperServerImpl.java
```diff
@@ -158,7 +159,7 @@ public WordRenderArtifactWrapperServerImpl(Object artifact, BranchSpecification
                   .segment( "Branch Specification Branch", branchSpecification.getBranchId().getIdString() )
                   .toString();
 
-      this.branchToken = BranchToken.create(artifactBranchToken.getId(), artifactBranchToken.getName(), branchSpecification.getViewId() );
```
lucianovaglienti, line 161: Same as above

plugins/org.eclipse.osee.define/src/org/eclipse/osee/define/operations/publisher/publishing/WordRenderArtifactWrapperServerImpl.java
```diff
@@ -950,7 +951,7 @@ public void setBranchView(BranchSpecification branchSpecification) {
                   .toString();
       //@formatter:on
       this.branchToken = BranchToken.create(branchSpecification.getBranchId().getId(), this.branchToken.getName(),
```
lucianovaglienti, line 953: and above

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/ds/QueryData.java
```diff
@@ -242,6 +244,18 @@ public void addCriteria(Criteria criteria) {
       getLastCriteriaSet().add(criteria);
    }
 
+   public QueryBuilder removeCriteria(Criteria criteria) {
```
lucianovaglienti, line 247: This seems mildly dangerous especially considering the query builder isn't necessarily thread-safe. @audrey-denk

web/apps/osee/src/app/ple/artifact-explorer/lib/components/hierarchy/artifact-hierarchy-panel/artifact-hierarchy-panel.component.ts
```diff
@@ -121,16 +132,57 @@ export class ArtifactHierarchyPanelComponent {
...
-			)
-		),
-		{
-			initialValue: false,
-		}
+			),
+			startWith(false)
+		);
+
+	displayWorkflowButtons = toSignal(
```
lucianovaglienti, line 154: Some simplifications(and maybe some correctness fixes):

```ts
private _branchCategories$ = toSignal(this.branchCategories, []);
categories = computed(()=>{this._branchCategories$().map(x->x.name)});
hasAts = computed(()=>this.categories().some(x=>x=='ATS');
isPle = computed(()=>this.categories().some(x=>x=='PLE'); //note: this is not needed
//do similar for workflow token valid
isBaseline = computed(()=> this.branchType() ==='baseline');
displayWorkflowButtons = computed(()=> workflowTokenValid() || (hasAts() && isBaseline()) );
showCreateBranch = computed(()=> !displayWorkflowButtons() && isBaseline()); 
```

The show create branch should show up for non PLE branches. The simplified version should only show up if !ATS, but the logic for baseline vs working should still be the same.

Realistically for the next part it might need a bit more granularity.

lucianovaglienti (approved):
A couple of nits... I think your guard is a little over-restrictive.

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/ds/QueryData.java
audrey-denk: If this was added just for the sake of getting the count, I don't think we need it at all.  It doesn't make the count any different if you include branch categories or not.  Was this just to improve time?  Sorry, memory is fuzzy.

web/apps/osee/src/app/ple/artifact-explorer/lib/components/hierarchy/artifact-hierarchy-panel/artifact-hierarchy-panel.component.ts
jadenpuckett: great suggestion I like how this looks

audrey-denk (commented):
Maybe we need to talk through the count thing... I remember talking about removing the include for the count, but I think it would be best as its own query instead of re-using the query that gets back the branches.

PR #302, opened 2025-05-22
feature[TW28559]: ICD - Add double buffer attribute to message level

audrey-denk:
Change-Id: I818af853d67a48440adaf380bed8c5eb484ccb5f


jadenpuckett (approved):

web/apps/osee/src/app/ple/messaging/message-tables/lib/dialogs/add-message-dialog/add-message-dialog.component.html
```diff
@@ -74,6 +74,19 @@ <h1 mat-dialog-title>Create New Message</h1>
 				data-cy="field-message-number" />
 		</mat-form-field>
 		<br />
+		<mat-slide-toggle
+			[(ngModel)]="doubleBuffer"
+			name="double_buffer"
+			[labelPosition]="doubleBuffer() ? 'after' : 'before'"
+			data-cy="field-double-buffer"
```
lucianovaglienti, line 81: I know this is probably copy & paste from elsewhere in the dialog... but can we use data-testid pattern instead since we aren't using cypress anymore?

plugins/org.eclipse.osee.mim/src/org/eclipse/osee/mim/internal/MimIcdGenerator.java
```diff
@@ -1021,8 +1023,9 @@ private void writeStructureSummarySheet(ExcelWorkbookWriter writer, List<Artifac
                   color = getCellColor(structReadable, subMessage.getArtifactReadable(), CoreAttributeTypes.InterfaceSubMessageNumber.getId());
                   color = color.equals(CELLSTYLE.NONE) ? getHeaderStructureCellColor(structureInfo.isAdded) : color;
                   writer.writeCell(rowIndex, 11, structureInfo.subMsgNum, color);
-                  writer.writeCell(rowIndex, 12, structureInfo.taskfile, getHeaderStructureCellColor(structureInfo.isAdded));
-                  writer.writeCell(rowIndex, 13, structureInfo.description, getHeaderStructureCellColor(structureInfo.isAdded), CELLSTYLE.WRAP);
+                  writer.writeCell(rowIndex, 12, structureInfo.doubleBuffered,color);
```
lucianovaglienti, line 1026: Found what I was looking for the other day... You should do something like the IcdElementIndex(https://github.com/eclipse-osee/org.eclipse.osee/blob/08affb44ff6e4092d70ec585f573b39ecf41545a/plugins/org.eclipse.osee.mim/src/org/eclipse/osee/mim/types/IcdElementIndex.java) so you can keep the column indexes lined up easier.

web/apps/osee/src/app/ple/messaging/shared/services/ui/header.service.ts
```diff
@@ -269,6 +269,11 @@ export class HeaderService {
 			description: 'Type of Message',
 			humanReadable: 'Type',
 		},
+		{
+			header: 'interfaceMessageDoubleBuffer',
+			description: 'TBD',
```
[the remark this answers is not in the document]
audrey-denk: Yes, I didn't notice this before.

PR #358, opened 2025-08-07
feature[TW28792]: Fix enumeration types adding values base on query r…

dondunne:
…esults

Change-Id: Ib35d27373fc6f6d5d8ab854d82472c7f9955966d


audrey-denk (dismissed):

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/data/AttributeTypeEnum.java
```diff
@@ -148,9 +149,14 @@ public T valueFromStorageString(String storedValue) {
             return enumToken;
          }
       }
-      T enumeration = enumTokens.get(0).clone(Long.valueOf(enumTokens.size()));
+      T enumeration = enumTokens.get(0).clone(Long.valueOf(enumTokens.size() + dbLoadedEnumTokens.size()));
```
rbrooks77, line 152: also need to look for stored value in dbLoadedEnumTokens (before creating a new one) to avoid repeatedly creating new enums for the same storedValue found only in the db

dondunne: done

rbrooks77 (approved):
Looks good.

rbrooks77 (approved):
Great!

PR #363, opened 2025-08-13
bug[TW28868]: PRs getting Priority and Change Type set to None and null

dondunne:
Change-Id: I048b1506bac1ca6253a5b4e18d04b2abb88cfb09

### DIFF
--- a/plugins/org.eclipse.osee.ats.api/src/org/eclipse/osee/ats/api/team/Priorities.java
+++ b/plugins/org.eclipse.osee.ats.api/src/org/eclipse/osee/ats/api/team/Priorities.java
@@ -56,6 +56,16 @@ public class Priorities extends OseeEnum {
       values.add(this);
    }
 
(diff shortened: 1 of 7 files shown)


plugins/org.eclipse.osee.ats.api/src/org/eclipse/osee/ats/api/team/Priorities.java
```diff
@@ -56,6 +56,16 @@ protected Priorities(String name, String description) {
       values.add(this);
    }
 
+   @JsonIgnore
+   public boolean isNone() {
+      return this.equals(Priorities.None);
+   }
+
+   @JsonIgnore
+   public boolean isNotNone() {
```
Grumb1e, line 65: This probably doesn't need to be a method, but doesn't really hurt anything either. : )

Grumb1e (approved):

PR #366, opened 2025-08-14
feature[TW28850]: Improve artifact link handling and table cell conte…

jadenpuckett:
…nt handling in Markdown conversion tool


Grumb1e (approved):

plugins/org.eclipse.osee.define/src/org/eclipse/osee/define/rest/importing/parsers/WordTemplateContentToMarkdownContentConverter.java
david-w-miller, line 1: At the last line, instead of adding the Strings together, just continue to append - 
logHeader.append("\n").append(errorLog.toString());
return logHeader.toString();

david-w-miller (approved):

PR #374, opened 2025-08-28
bug[TW28895]: Safeguard against transactions with Sentinel author

dondunne:


plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionFactoryImpl.java
```diff
@@ -96,34 +97,30 @@ public CancellableCallable<Integer> purgeTransaction(Collection<? extends Transa
...
-
+      return createTransaction(branch, orcsApi.userService().getUser(), comment);
    }
 
    @Override
    public TransactionBuilder createTransaction(BranchId branch, UserId author, String comment) {
+      Conditions.assertNotNull(author, "Author can not be null");
+      Conditions.assertNotSentinel(author, "Author can not be Sentinel");
+      Conditions.assertNotNullOrEmpty(comment, "Comment must be specified");
+      if (!queryFactory.branchQuery().andId(branch).exists()) {
```
Grumb1e, line 108: Have we always required comments somewhere? Adding the requirement if it did not exist previously may break something.

dondunne: yes, always required, but not checked at the lowest level

jadenpuckett (approved):

Grumb1e (dismissed):

Grumb1e (approved):

PR #375, opened 2025-09-05
feature[TW28900]: Create rest call returning a type with specific relation

audrey-denk:
…ation

Change-Id: I8975ff51a3dce69d76733763bd1827d5201ed8d0

### DIFF
--- a/plugins/org.eclipse.osee.ats.core/src/org/eclipse/osee/ats/core/task/track/ScriptTaskTrackingOperation.java
+++ b/plugins/org.eclipse.osee.ats.core/src/org/eclipse/osee/ats/core/task/track/ScriptTaskTrackingOperation.java
@@ -11,6 +11,7 @@
 package org.eclipse.osee.ats.core.task.track;
 
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Date;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/ds/OptionsUtil.java
```diff
@@ -239,6 +240,14 @@ public static boolean getNoLoadRelations(Options options) {
       return options.getBoolean(NO_LOAD_RELATIONS);
    }
 
+   public static void setFollowNamesOnly(Options options, boolean enabled) {
```
lucianovaglienti, line 243: If we are going to go through the work for doing this for Names, why don't we do this and store an AttributeId instead? i.e. ONLY_FOLLOW_ATTRIBUTE = "only.follow.attribute"

plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/sql/SelectiveArtifactSqlWriter.java
```diff
@@ -453,6 +454,10 @@ private String writeAttsCommonTableExpression(String artWithAlias) {
          writeEqualsAnd(attTxsAlias, txdAlias, "transaction_id");
       }
       writeTxBranchFilter(attTxsAlias);
+      if (OptionsUtil.getFollowNamesOnly(getOptions())) {
+         write(
+            " and case when top = 0 then " + CoreAttributeTypes.Name.getIdString() + " else att.attr_type_id end = att.attr_type_id");
```
lucianovaglienti, line 459: Same comment as above...I don't think we need to hardcode this to name

audrey-denk: same reply.

david-w-miller (approved):

PR #383, opened 2025-09-17
bug[TW28897]: Simplify client user loading

dondunne:
Change-Id: I76840f22c4034ee663817bc4b2b08d9c87a8b79a


audrey-denk (approved):

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/UserManager.java
```diff
@@ -39,63 +38,48 @@ public final class UserManager {
...
 
    public static void releaseUser() {
-      getUserAdmin().releaseCurrentUser();
+      getUserService().clearCaches();
    }
 
    public static void clearCache() {
-      getUserAdmin().reset();
+      getUserService().clearCaches();
    }
```
Grumb1e, line 76: These two methods are now doing the same thing. Was that intentional? Do we need both?

dondunne: Same, but I'd like to keep the separation cause the uses make it more descriptive of what we're trying to do.  UserManager will eventually (soon) go away, so this will be cleaned up then.

Grumb1e (approved):

PR #387, opened 2025-09-19
feature[TW28941]: Convert IDE to load users from server - remove User…

dondunne:
…Manager

Change-Id: I52aea6385c3de4d97683bc827b17985fc24f12e1

### DIFF
--- a/plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/enums/CoreArtifactTypes.java
+++ b/plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/enums/CoreArtifactTypes.java
@@ -338,9 +338,15 @@ public interface CoreArtifactTypes {
       .exactlyOne(SeverityCategory, SeverityCategory.Unspecified)
       .zeroOrOne(SubjectMatterExpert));
 
-   ArtifactTypeToken DesignDescriptionMsWord = osee.add(osee.artifactType(810L, "Design Description - MS Word", false, MsWordTemplate, AbstractHeading));
+   ArtifactTypeToken DesignDescriptionMsWord = osee.add(osee.artifactType(810L, "Design Description - MS Word", false, MsWordTemplate)
+      .zeroOrOne(DoorsHierarchy)
+      .any(DoorsId)
+      .any(DoorsModId));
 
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/access/UserServiceImpl2.java
```diff
@@ -0,0 +1,321 @@
...
+import org.eclipse.osee.framework.skynet.core.internal.ServiceUtil;
+import org.eclipse.osee.orcs.rest.model.DatastoreEndpoint;
+
+/**
+ * Client Implementation with OseeUser loaded from server
+ *
+ * @author Donald G. Dunne
+ */
+
+public class UserServiceImpl2 implements UserService2 {
```
Grumb1e, line 53: I would suggest more meaningful names for these than "2". Perhaps RemoteUserService, ApiUserService or something else indicate that it talks with the server. Additionally, it would probably be worth adding a little Java doc to differentiate this from the legacy. 

You could potentially just call this UserService and rename the old one to UserServiceLegacy.

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/access/UserServiceImpl2.java
```diff
@@ -0,0 +1,321 @@
...
+      IUserGroup group = getUserGroupOrNull(userGroup);
+      if (group != null) {
+         return group;
+      } else {
+         throw new OseeArgumentException("parameter must be artifact");
+      }
+   }
+
+   @Override
+   public IUserGroup getUserGroupOrNull(IUserGroupArtifactToken userGroup) {
```
Grumb1e, line 198: I believe optionals are now preferred over null. Null works though.

plugins/org.eclipse.osee.framework.skynet.core/src/org/eclipse/osee/framework/skynet/core/internal/OseeUserImpl.java
```diff
@@ -0,0 +1,127 @@
+/*
+ * Created on Sep 17, 2025
+ *
+ * PLACE_YOUR_DISTRIBUTION_STATEMENT_RIGHT_HERE
+ */
```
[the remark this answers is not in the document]
dondunne: done

PR #395, opened 2025-09-24
feature[TW28941]: Convert IDE to load users from server - update User…

dondunne:
…ServiceImpl

Change-Id: I88e888c659aec7525aa24f6fa5cd5602f75d2a7f


plugins/org.eclipse.osee.ats.core/src/org/eclipse/osee/ats/core/access/AtsArtifactChecks.java
```diff
@@ -247,23 +248,22 @@ private void checkAtsWorkDefinitions(boolean isAtsAdmin, AtsApi atsApi, Collecti
 
    private void checkUsers(boolean isAtsAdmin, AtsApi atsApi, Collection<ArtifactToken> artifacts,
       XResultData results) {
-      Set<UserToken> users = new HashSet<>();
       for (ArtifactId art : artifacts) {
-         if (art instanceof UserToken) {
-            if (!isAtsAdmin && !atsApi.isInTest()) {
-               results.error("Only Admins can delete users");
-               return;
+         if (atsApi.getStoreService().isOfType(art, CoreArtifactTypes.User)) {
```
david-w-miller, line 252: I couldn't tell if this is making a rest call for each user to check its type?

Grumb1e (approved):

PR #399, opened 2025-09-29
feature[TW26095]: Integrate OSEE Markdown to HTML renderer with web A…

jadenpuckett:
…rtifact Explorer Markdown editor

### DIFF
--- a/plugins/org.eclipse.osee.framework.core/OSEE-INF/markdown/markdownToHtmlStyles.css
+++ b/plugins/org.eclipse.osee.framework.core/OSEE-INF/markdown/markdownToHtmlStyles.css
@@ -3,7 +3,7 @@ table {
 }
 
 table, th, td {
-	border: 1px solid black;
+	border: 1px solid;
 	border-collapse: collapse;
 	padding: 5px;
 	page-break-inside: avoid;
--- a/web/apps/osee/src/app/shared/components/markdown-editor/markdown-editor.component.html
+++ b/web/apps/osee/src/app/shared/components/markdown-editor/markdown-editor.component.html
@@ -10,105 +10,70 @@
  * Contributors:
  *     Boeing - initial API and implementation
  -->
-
-<div class="tw-flex tw-flex-row tw-gap-4 tw-p-4">
-	<button
-		mat-icon-button
-		matTooltip="Undo"
-		(click)="undo()">
-		<div>
-			<mat-icon> undo </mat-icon>
-		</div>
-	</button>
-	<button
-		mat-icon-button
-		matTooltip="Redo"
-		(click)="redo()">
-		<div>
-			<mat-icon> redo </mat-icon>
-		</div>
-	</button>
-	<button
-		mat-icon-button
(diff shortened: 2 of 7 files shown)


web/apps/osee/src/app/shared/components/markdown-editor/markdown-editor.component.ts
```diff
@@ -79,6 +72,34 @@ export class MarkdownEditorComponent {
 	maxHistory = signal(100);
 	mdExamples = mdExamples;
 
+	artExpHttpService = inject(ArtifactExplorerHttpService);
+	domSanitizer = inject(DomSanitizer);
+
+	// Markdown Preview
+
+	toggleRefreshPreview() {
```
lucianovaglienti, line 80: This isn't true. toggleRefreshPreview will never "toggle"

web/apps/osee/src/app/shared/components/markdown-editor/markdown-editor.component.ts
```diff
@@ -79,6 +72,34 @@ export class MarkdownEditorComponent {
...
+		this.refreshPreview.set(true);
+	}
+	refreshPreview = signal<boolean>(true);
+	mdPreview = toSignal(
```
lucianovaglienti, line 84: We shouldn't be setting a value inside a switchmap like this. Use an rxjs operator like repeatWhen or retryWhen if you need to trigger the request on a button press.

web/apps/osee/src/app/shared/components/markdown-editor/markdown-editor.component.html
```diff
@@ -10,105 +10,80 @@
...
-				class="tw-text-base">
-			</markdown>
-		</div>
+		</button>
 	</div>
 </div>
+<!-- Editor -->
+<div class="tw-flex tw-size-full tw-border tw-border-solid">
+	<!-- Markdown Editor -->
+	<mat-form-field
```
lucianovaglienti, line 68: Just a note: probably want to make this component take in the attribute type that the "parent" component uses, so when Carl gets persistence finished, he can update here too.

plugins/org.eclipse.osee.framework.core/OSEE-INF/markdown/markdownToHtmlStyles.css
```diff
@@ -3,7 +3,7 @@ table {
 }
 
 table, th, td {
-	border: 1px solid black;
+	border: 1px solid;
 	border-collapse: collapse;
 	padding: 5px;
 	page-break-inside: avoid;
```
lucianovaglienti, line 9: Just a q: Can we update publishing at some point to just use the generated CSS that we get from tailwind/ @angular/material?

jadenpuckett: write the action!

lucianovaglienti (approved):

audrey-denk (approved):

PR #531, opened 2026-03-19
feature[TW29360]: Create Angular publishing page

david-w-miller:


web/apps/osee/src/app/ple/book-explorer/book-explorer.component.ts
```diff
@@ -0,0 +1,678 @@
...
+	private readonly dropdownContentEffect = effect(() => {
+		const cfg = this.config();
+		const branchId = this.currentBranchId;
+
+		if (!cfg || !branchId) return;
+
+		this.loadDropdownContent(cfg);
+	});
+
+	get currentBranchId(): string {
```
lucianovaglienti: Why is this needed?

web/apps/osee/docs/developer-notes/ple/publish/publish-launcher.md
```diff
@@ -0,0 +1,399 @@
...
+- `{email}`
+
+This allows the UI to automatically render an **Email Addresses** input field only for workflows that require it.
+
+#### 4. Dropdowns can be static or dynamic
+
+Dropdowns are defined per tab and support two modes:
+
+- **Static options** via `options`
+- **Dynamic options** via `contentApi`
```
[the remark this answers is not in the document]
david-w-miller: Since the rest API is selected by the page creator, I wanted to keep it simple. The current API set for publishing does not need pagination. One solution is when developing pages for this tool, write specific tabs and rest APIs that filter so that pagination is never needed.

web/apps/osee/src/app/ple/book-explorer/book-explorer.component.ts
```diff
@@ -0,0 +1,678 @@
+/*********************************************************************
```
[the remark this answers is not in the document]
david-w-miller: I re-wrote the component, and attempted a split up. It should be more declarative now also.

web/apps/osee/src/app/ple/book-explorer/book-explorer.component.ts
```diff
@@ -0,0 +1,678 @@
...
+	value?: string;
+}
+
+interface TabCheckbox {
+	key: string;
+	label: string;
+	default?: boolean;
+}
+
+interface TargetApi {
```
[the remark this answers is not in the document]
david-w-miller: I reviewed the MIM reports, in this case I was looking for something a bit simpler that I could define new pages without changing code.

web/apps/osee/src/app/ple/book-explorer/book-explorer.component.ts
```diff
@@ -0,0 +1,678 @@
...
+		const request =
+			dd.contentApi.method === 'GET'
+				? this.http.get<DropdownApiItem[]>(url)
+				: this.http.post<DropdownApiItem[]>(url, {});
+
+		request.subscribe((items) => {
+			this.dropdownOptions[tab.key][dd.key] =
+				this.mapDropdownItems(items);
+		});
+	}
```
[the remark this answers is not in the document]
david-w-miller: I changed this to be more in line with your comment, it should be declarative now.

web/apps/osee/src/app/ple/book-explorer/book-explorer.component.ts
```diff
@@ -0,0 +1,678 @@
...
+		return this.getDropdownOptionsFor(tab, dd).length > 10;
+	}
+
+	private tryParseResponse(response: string): unknown {
+		try {
+			return JSON.parse(response);
+		} catch {
+			return response;
+		}
+	}
```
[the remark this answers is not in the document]
david-w-miller: removed/re-written

web/apps/osee/src/app/ple/book-explorer/book-explorer.component.ts
```diff
@@ -0,0 +1,678 @@
...
+
+	private showResults(content: unknown) {
+		this.dialog.open(BookExplorerResultDialogComponent, {
+			data: { content },
+			maxWidth: '95vw',
+		});
+	}
+
+	doWork(tab: BookExplorerTabConfig) {
```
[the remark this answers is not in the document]
david-w-miller: It was an homage to old OSEE, I removed it.

web/apps/osee/src/app/ple/book-explorer/book-explorer.component.ts
```diff
@@ -0,0 +1,678 @@
...
+		const branchId = this.currentBranchId;
```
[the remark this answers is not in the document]
david-w-miller: I tried updating to all new Angular best practice - let me know how close it is now.

web/apps/osee/src/app/ple/ple.component.html
```diff
@@ -46,4 +46,14 @@
 			AcTra
 		</a>
 	</div>
+	<div
+		class="tw-grid tw-max-w-[822px] tw-grid-cols-landing tw-justify-center tw-gap-7">
```
lucianovaglienti: How does this look if we expand to 3 per line? Or can we make these rows a bit more dynamic so the buttons shrink/grow a little bit? Just to me seems a bit yucky that we are doing a fixed px number here(didn't realize we were doing it this way).

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.types.ts
```diff
@@ -0,0 +1,72 @@
...
+	readonly description?: string;
+	readonly instructions?: readonly string[];
+	readonly dropdowns: readonly TabDropdown[];
+	readonly checkboxes: readonly TabCheckbox[];
+	readonly targetApi: TargetApi;
+	readonly artifact?: string;
+};
+
+export type PublishLauncherConfig = {
+	readonly title?: string;
```
lucianovaglienti: In what scenario would a title be undefined?

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.types.ts
```diff
@@ -0,0 +1,72 @@
...
+export type TabCheckbox = {
+	readonly key: string;
+	readonly label: string;
+	readonly default?: boolean;
+};
+
+export type PublishLauncherTabConfig = {
+	readonly key: string;
+	readonly label: string;
+	readonly description?: string;
```
lucianovaglienti: I would recommend these to always return from their api and in the web front end here(i.e. description: string; instructions:string[])

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.types.ts
```diff
@@ -0,0 +1,72 @@
...
+export type TargetApi = {
+	readonly method: 'GET' | 'POST';
+	readonly url: string;
+	readonly button?: string;
+};
+
+export type TabDropdown = {
+	readonly key: string;
+	readonly label: string;
+	readonly value?: 'required';
```
lucianovaglienti: Is the type here really 'required'|undefined? That seems like a bad way to handle an option.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
```diff
@@ -0,0 +1,128 @@
...
+			switchMap(() =>
+				this.httpService
+					.getConfig()
+					.pipe(shareReplay({ bufferSize: 1, refCount: true }))
+			)
+		),
+		{ initialValue: undefined }
+	);
+
+	constructor() {
```
lucianovaglienti: You have probably already fixed locally:
move the effect to a variable

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-http.service.ts
```diff
@@ -0,0 +1,86 @@
...
+} from './publish-launcher.types';
+
+@Injectable({
+	providedIn: 'root',
+})
+export class PublishLauncherHttpService {
+	private readonly http = inject(HttpClient);
+	private readonly uiService = inject(UiService);
+
+	getConfig(): Observable<PublishLauncherConfig> {
```
lucianovaglienti: If you are only subscribing to this at the top level, I recommend converting this to an httpResource or an fn(signal) -> HttpResource.
If the passed in signal value === '-1', trigger undefined instead of url

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
```diff
@@ -0,0 +1,128 @@
...
+})
+export class PublishLauncherComponent {
+	branchId = input<string>('');
+	branchType = input<BranchType>('');
+
+	private readonly uiService = inject(UiService);
+	private readonly branchRouteState = inject(BranchRoutedUIService);
+	private readonly httpService = inject(PublishLauncherHttpService);
+
+	protected readonly loadingGlobal = toSignal(this.uiService.isLoading, {
```
lucianovaglienti: This comment goes along with the comment on the service, if you switch over to an http resource/ fn(signal) -> httpResource, loading states can be provided by your resource.

```ts
private _branchId = toSignal(this.uiService.id, '-1');
protected readonly configResource = publishingService.getConfig(this._branchId);
protected readonly isLoadingConfig = configResource.isLoading();
protected readonly config = configResource.value();
```

Also this should probably just be in a service like so:

publishing.service.ts:
```ts
private readonly uiService = inject(UiService);
private readonly httpService = inject(PublishLauncherHttpService);
private _branchId = toSignal(this.uiService.id, '-1');
readonly configResource = publishingService.getConfig(this._branchId);
readonly isLoadingConfig = configResource.isLoading();
readonly config = configResource.value();
```

publish-launcher.component.ts:
```ts
branchId = input<string>('');
branchType = input<BranchType>('');
private readonly uiService = inject(UiService);
private publishingService = inject(PublishingService);
private _updateBranches = effect(() => {
			const branchId = this.branchId();
			const branchType = this.branchType();
			if (branchId) this.uiService.idValue = branchId;
			if (branchType) this.uiService.typeValue = branchType;
		});
protected config = publishingService.config;
protected isLoading = publishingService.isLoadingConfig;
```

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
```diff
@@ -0,0 +1,128 @@
...
+		BranchPickerComponent,
+		PublishLauncherTabComponent,
+	],
+	template: `
+		<div class="tw-p-4">
+			<div class="tw-mb-4">
+				<osee-branch-picker />
+			</div>
+
+			@if (!hasBranch()) {
```
lucianovaglienti: This can be cleaned up a lot:

```html
@if (config.isLoading()){
<div class="tw-mb-4 tw-rounded tw-border tw-border-blue-300 tw-bg-blue-50 tw-p-3 tw-text-blue-900 dark:tw-border-blue-700 dark:tw-bg-blue-950 dark:tw-text-blue-200">
						Loading Publishing Configuration...
					</div>
} @else if(config().title){
<h1 class="tw-mb-6 tw-text-2xl tw-font-bold">
						{{ config.title }}
					</h1>
}

```
Remove the select a branch part, that's managed by the branch picker.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
```diff
@@ -0,0 +1,128 @@
...
+					</div>
+				}
+
+				<mat-tab-group>
+					@for (tab of config.tabs; track tab.key) {
+						<mat-tab [label]="tab.label">
+							<div class="tw-p-4">
+								<osee-publish-launcher-tab
+									[tab]="tab"
+									[branchId]="currentBranchId()"
```
lucianovaglienti, line 66: branchId and branchType don't need to propagated down, as it's in the uiService.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,458 @@
...
+	FormState,
+	TabDropdown,
+} from './publish-launcher.types';
+
+@Component({
+	selector: 'osee-publish-launcher-tab',
+	changeDetection: ChangeDetectionStrategy.OnPush,
+	imports: [
+		ReactiveFormsModule,
+		MatButtonModule,
```
lucianovaglienti: Import the non-Module equivalents

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
```diff
@@ -0,0 +1,128 @@
...
+import { filter, shareReplay, switchMap } from 'rxjs';
+import { PublishLauncherHttpService } from './publish-launcher-http.service';
+import { PublishLauncherTabComponent } from './publish-launcher-tab.component';
+import type { BranchType } from './publish-launcher.types';
+
+@Component({
+	selector: 'osee-publish-launcher',
+	changeDetection: ChangeDetectionStrategy.OnPush,
+	imports: [
+		MatTabsModule,
```
lucianovaglienti: Import the non-module equivalents.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,458 @@
...
+	private readonly httpService = inject(PublishLauncherHttpService);
+
+	private readonly dropdownOptions = signal<DropdownState>({});
+	private readonly dropdownFilters = signal<FilterState>({});
+
+	protected readonly loading = toSignal(this.uiService.isLoading, {
+		initialValue: 'false',
+	});
+
+	readonly form = computed(() => this.buildForm());
```
lucianovaglienti: You might want to try the new experimental forms with Angular 21.
https://angular.dev/essentials/signal-forms

You can make life easier using:
https://angular.dev/guide/forms/signals/validation#validation-of-array-items
https://angular.dev/guide/forms/signals/validation#custom-validation-rules

There's a lot of mutable funkiness going on in here that I don't quite understand.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,458 @@
...
+
+	constructor() {
+		effect(() => {
+			const tab = this.tab();
+			const branchId = this.branchId();
+			if (tab && branchId) {
+				this.loadDropdownData(tab, branchId);
+			}
+		});
+	}
```
lucianovaglienti: Move this to be a non-constructor effect.

Edit: Smarter more declarative way to do this:

```ts
private id = toSignal(this.uiService.id,'-1');
readonly dropdowns = computed(()=>this.tab().dropdowns);
private dropDowns$ = toObservable(this.dropdowns);
private dropDownIter = from(this.dropdowns$);
private dropdownData$ = this.dropdownsIter.pipe(
concatMap(dropdown=>{
if (dropdown.contentApi){
         const url = this.resolveUrlTemplate(
					dropdown.contentApi.url,
					this.tab(),
					{},
					this.id()
				);
         return this.httpService
					.getDropdownItems(dropdown.contentApi.method, url)
					.pipe(
						map((items) => ({
							dropdownKey: dropdown.key,
							options: items.map((item) => ({
								id: item.id,
								label: item.name,
							})),
						}))
					);
    }
    return of({
     dropdownKey: dropdown.key,
     [dropdownKey]: dropdown.options
    })
})
);
protected dropdownData = toSignal(this.dropdownData$, {initialValue:[]});
```

I probably didn't get this 100% correct because I don't understand this code 100%, and it's been quite a few months since I've used concatMap/mergeMap :)

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,458 @@
...
+							</li>
+						}
+					</ul>
+				</div>
+			}
+
+			<form
+				[formGroup]="form()"
+				class="tw-space-y-4">
+				@if (requiresEmail()) {
```
lucianovaglienti: You should use [required]="requiresEmail()" instead of @ if(requiresEmail())

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,458 @@
...
+	): string {
+		const branch = branchId || this.branchId();
+		const artifact = tab.artifact || '';
+
+		const replacements: Record<string, string> = {
+			branch,
+			branchId: branch,
+			artifact,
+			artifactId: artifact,
+			email: String(formValue['email'] || '').trim(),
```
lucianovaglienti, line 330: I think with a sensible model/pass more reasonable things along you can just make this :
email: (email(or model().email)|| '').trim()

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,458 @@
...
+		for (const dropdown of tab.dropdowns) {
+			const value = formValue[dropdown.key];
+			if (value !== null && value !== undefined && value !== '') {
+				replacements[dropdown.key] = String(value).trim();
+			} else if (!this.isDropdownRequired(dropdown)) {
+				replacements[dropdown.key] = '-1';
+			}
+		}
+
+		return url.replace(/\{([^}]+)\}/g, (match, key: string) => {
```
lucianovaglienti: This regex should be extracted out to a constant, with some documentation to explain what it's doing.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,458 @@
...
+
+		const replacements: Record<string, string> = {
+			branch,
+			branchId: branch,
+			artifact,
+			artifactId: artifact,
+			email: String(formValue['email'] || '').trim(),
+		};
+
+		for (const dropdown of tab.dropdowns) {
```
lucianovaglienti: This feels like something that could live in an computed.

like 
```ts
private _calculatedReplacements = computed(()=>{
const replacements: Record<string, string> = {
			id(),
			branchId: id(),
			tab().artifact,
			artifactId: tab().artifact,
			email: form().email().value() || '').trim(),
};
for (const dropdown of dropdowns()){
const dropdownKey = form()[dropdown.key]();
if (dropdownKey.valid() && dropdownKey.touched()){
   replacements[dropdown.key] = dropdownKey.value().trim();
} else if (!dropdownKey.required()){
   replacements[dropdown.key] = '-1';
}
return replacements
}

});

private resolveUrlTemplate(url:string){
   const replacements = this._calculatedReplacements();
    return url.replace(/\{([^}]+)\}/g, (match, key: string) => {
        const value = replacements[key];
        return value !== undefined ? encodeURIComponent(value) : match;
    });
}
```
Now that I think a little bit more about it, it's probably just a computed.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,458 @@
...
+
+	readonly form = computed(() => this.buildForm());
+
+	readonly requiresEmail = computed(() =>
+		this.tab().targetApi.url.includes('{email}')
+	);
+
+	readonly publishTooltip = computed(() => {
+		if (!this.form().valid) return 'Please fill in all required fields';
+		if (this.loading() === 'true')
```
lucianovaglienti: Use the signal off of httpResources if possible

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,458 @@
...
+		this.dropdownFilters.update((state) => ({
+			...state,
+			[this.tab().key]: {
+				...state[this.tab().key],
+				[dropdownKey]: value || '',
+			},
+		}));
+	}
+
+	executePublish(): void {
```
lucianovaglienti, line 409: Not going to leave as detailed of a note here, but I think we can clean this up.
The publish should ideally just be triggered as a result of a couple signals changing (aka trigger an fn(signal)-> httpResource) that does undefined when values aren't set.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,458 @@
...
+
+				<div class="tw-pt-4">
+					<button
+						mat-flat-button
+						color="primary"
+						class="dark:tw-bg-primary-400 dark:tw-text-black"
+						[disabled]="!form().valid || loading() === 'true'"
+						(click)="executePublish()"
+						[matTooltip]="publishTooltip()">
+						{{ tab.targetApi.button || 'Do Work' }}
```
lucianovaglienti: Can we call this 'Execute Publish' or something like that instead, Do Work gives me PTSD from all the stuff present in the client and server...and ideally I'd like to be able to grep for that and not run into web code.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,458 @@
...
+						{{ checkbox.label }}
+					</mat-checkbox>
+				}
+
+				<div class="tw-pt-4">
+					<button
+						mat-flat-button
+						color="primary"
+						class="dark:tw-bg-primary-400 dark:tw-text-black"
+						[disabled]="!form().valid || loading() === 'true'"
```
lucianovaglienti: Wait, you are using reactive forms, but your form is a signal???

Just use signal forms, stick to one system.

I'll comment on the rest of the HTML once the rest of the ts gets touched up a bit...

web/apps/osee/src/app/layout/lib/navigation/top-level-navigation/top-level-navigation-structure.ts
```diff
@@ -62,6 +62,19 @@ export const navigationStructure: navigationElement[] = [
...
 				children: [],
 			},
+			{
+				label: 'Publishing',
+				cypressLabel: 'publishing',
+				pageTitle: 'OSEE - Publish Launcher',
+				isDropdown: false,
+				isDropdownOpen: false,
+				requiredRoles: [],
+				routerLink: '/ple/publishing',
```
lucianovaglienti, line 72: Just a Q: are we intending for publishing to be under ple? Or should it be /publishing?

david-w-miller: I put it there because the artifact explorer is there, and because the publish should be thought of as a publish in a view.

web/apps/osee/src/app/ple/ple.component.html
david-w-miller: Actions Taken :

Read ple.component.html — found 3 separate grid 'div's each with tw-max-w-[822px] tw-grid-cols-landing, buttons had fixed tw-w-landing-button (255px)
Checked tailwind.config.js — confirmed landing: 'repeat(auto-fit, minmax(255px, 1fr))' and 'landing-button': '255px'
Applied changes via fsReplace:
Merged all 3 grid divs into a single grid container
Removed tw-w-landing-button from all 5 buttons so they grow via 1fr
Changed container from tw-max-w-[822px] to tw-w-full tw-max-w-4xl tw-px-4 (896px, enough for 3 cols)

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.types.ts
david-w-miller: If the Publishing Json doesn't include a title, this would handle that case.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.types.ts
david-w-miller: I am OK with requiring the description and instructions - I will update the code.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.types.ts
david-w-miller: It seemed natural to me - if not required, the JSON doesn't have to have a setting. If it is required, then JSON is marked with value is required. I can update the JSON with value is optional fields everywhere and change the code it you think that is best.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
david-w-miller: go it

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.types.ts
lucianovaglienti: I think my point here is that title should probably always be defined. If you parse JSON and it doesn't have a title, I think it's an error.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.types.ts
lucianovaglienti: I think it's probably more natural to have required: boolean or required: 'true'|'false', and if the json parse doesn't include it, to default it to false.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
david-w-miller: This is a long running publish launcher - it is best if they don't accidentally publish. If someone kicks of five colorbook publishes changing options around the server will break. Possibly I should put an "Are you absolutely sure" box on the execute publish.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
david-w-miller: I think I got them all now.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
david-w-miller: I tried making it all signal forms, but it broke the branch widget somehow. I backed that out and made a minor reactive compute update.

lucianovaglienti: That means you likely converted too much, the branch selector is outside the form

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
lucianovaglienti: the button can just publish a new version of a publish signal and we can trigger off that.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.types.ts
david-w-miller: I agree - changed

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
david-w-miller: done

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
david-w-miller: done

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
david-w-miller: I ran into a timing issue between components with using uiService directly, so I am reluctant to change this.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
david-w-miller: done

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
```diff
@@ -0,0 +1,113 @@
...
+	computed,
+	effect,
+	inject,
+	input,
+} from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
+import { httpResource } from '@angular/common/http';
+import { MatTab, MatTabGroup } from '@angular/material/tabs';
+import { BranchPickerComponent } from '@osee/shared/components';
+import { BranchRoutedUIService, UiService } from '@osee/shared/services';
```
lucianovaglienti: Don't need to import BranchRoutedUIService, that's an internal impl for the branch picker typically.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
lucianovaglienti: If you are running into a timing issue, you are likely doing something wrong elsewhere.(i.e. you should just be using the branch id via :

```ts
private readonly uiService = inject(UiService);
private id = toSignal(this.uiService.id, {initialValue:''});
private _updateIdsOnVisit = effect(()=>{
if (this.id() !== this.branchId()){
this.uiService.idValue = this.branchId();
}
});
readonly hasBranch = computed(() => {
		const id = this._id();
		return !!id && id !== '0';
	});
```

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,443 @@
...
+		MatLabel,
+		MatError,
+		MatSelect,
+		MatOption,
+		MatCheckbox,
+		MatInput,
+		MatTooltip,
+	],
+	template: `
+		@if (tab(); as tab) {
```
lucianovaglienti: Just use @ let tab = tab(); Unless there's truly a condition tab() can result in undefined|null.

Also, probably need to call the two variables separate things so you don't get collisions.

jadenpuckett (dismissed):

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,443 @@
...
+
+	private readonly fb = inject(FormBuilder);
+	private readonly dialog = inject(MatDialog);
+	private readonly http = inject(HttpClient);
+	private readonly httpService = inject(PublishLauncherHttpService);
+
+	private readonly dropdownFilters = signal<FilterState>({});
+	protected readonly publishing = signal(false);
+
+	readonly form = signal<FormGroup>(this.fb.group({}));
```
lucianovaglienti: I want to emphasize here, this needs to either be a signal form or a reactive form. Not a signal containing a reactive form...this is messy.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,443 @@
...
+	});
+
+	readonly publishTooltip = computed(() => {
+		if (!this.form().valid) return 'Please fill in all required fields';
+		if (this.publishing())
+			return 'Please wait for current operation to complete';
+		return '';
+	});
+
+	private readonly rebuildFormEffect = effect(() => {
```
lucianovaglienti: See comment above, should probably just be signal form.
Also if you are using signals properly, you shouldn't need to run an effect as it'll run as soon as the signal values change to make a new value.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
```diff
@@ -0,0 +1,113 @@
...
+			<div class="tw-mb-4">
+				<osee-branch-picker />
+			</div>
+
+			@if (loadingGlobal()) {
+				<div
+					class="tw-mb-4 tw-rounded tw-border tw-border-blue-300 tw-bg-blue-50 tw-p-3 tw-text-blue-900 dark:tw-border-blue-700 dark:tw-bg-blue-950 dark:tw-text-blue-200">
+					Loading Publishing Configuration...
+				</div>
+			} @else if (config(); as config) {
```
lucianovaglienti, line 53: Use @ let

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
david-w-miller: done.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
david-w-miller: good point, changed to tabConfig, used @let instead.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
david-w-miller: During testing, the form was attempting to call the form loading rest APIs before the branch was selected. I only want the dropdown APIs called when the config is loaded and truthy. Without this guard. the block is attempted, even though the data that drives it is not available.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
david-w-miller: I put it into a computed signal instead of a signal + effect. Do you want this refactored to a signal form? That would mean replacing the FormGroup/ReactiveFormsModule approach entirely with Angular's signal-based forms API. Since that is labelled as experimental APIs, can we stick with  the current approach (reactive FormGroup inside a computed signal) as the safe choice?

lucianovaglienti: Angular's definition of "experimental" is significantly different than what you are used to. The APIs for signal forms might change slightly, but it's nothing to worry about. It is a significantly safer approach than having two forms of reactivity nesting each other. Historically, we have used experimental features for months before they became stable with no issues.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher.component.ts
lucianovaglienti: If you are switched over to the resource API, you can instead use @ if resource.loading()(isLoading? can't remember) to block.

jadenpuckett (dismissed):

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
david-w-miller: I switched this to a signal form.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
david-w-miller: This is now refactored as a signal form.

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,498 @@
...
+
+	private readonly formModel = signal<FormState>({ email: '' });
+
+	protected readonly publishForm = form(this.formModel, (path) => {
+		validate(path, (ctx) => {
+			const formValue = ctx.value();
+			const tabConfig = this.tab();
+			const errors = [];
+
+			if (this.requiresEmail()) {
```
lucianovaglienti: For this you can use the angular built in(https://angular.dev/essentials/signal-forms#validation-and-state  , https://angular.dev/guide/forms/signals/validation#email) or an apply to conditionally apply the angular built-in

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,498 @@
...
+								kind: 'email_invalid',
+								key: 'email',
+								invalidAddresses: invalid,
+							})
+						);
+					}
+				}
+			}
+
+			for (const dropdown of tabConfig.dropdowns) {
```
lucianovaglienti: Use required: https://angular.dev/guide/forms/signals/validation#required

web/apps/osee/src/app/ple/publish-launcher/publish-launcher-tab.component.ts
```diff
@@ -0,0 +1,498 @@
...
+								label: dropdown.label,
+							})
+						);
+					}
+				}
+			}
+
+			return errors.length > 0 ? errors : null;
+		});
+	});
```
lucianovaglienti, line 257: What I'm thinking this should look like:
```ts
protected dropdowns = computed(()=>this.tab().dropdowns);
protected requiredDropdowns = computed(()=>this.dropdowns().filter(x=>this.isDropdownRequired(x)));
//maybe we need to think of a good way to turn this into a computed, resource or linkedSignal
protected computedRequiredValidators =(schemaPath)=>{
    let validators = [];
    for (const dropdown of requiredDropdowns()) {
        validators.push(required(schemaPath[dropdown.key]));
    }
    return validators;
};

protected readonly publishForm = form(this.formModel,(schemaPath) =>{
    applyWhen(
        schemaPath.email,
       ({_value})=>this.requiresEmail(),
       (emailSchemaPath)=>{
            email(schemaPath.email)
       }
    ),
//this.requiresEmail()?email(schemaPath.email):()=>{},
    ...computedRequiredValidators(schemaPath)
//Even better: use applyEach and applyWhen in conjunction here to do the required addition
}
);
```

david-w-miller: Thanks for the suggestion and the detailed example — appreciate the thought here. We actually already use required() from @angular/forms/signals (for the email field) and have requiredDropdowns as a computed signal, so the reactive intent is covered.

The reason we went with a single validate(path, ...) block for the dropdown required checks rather than per-field required(schemaPath[dropdown.key]) calls is that the form shape is fully dynamic — the dropdowns come from a config input and the form model gets rebuilt at runtime. At schema construction time those keys don't reliably exist on the schema path yet, so iterating over them inside validate() lets us defer that check to validation time when the model is populated.

Using applyEach/applyWhen per-dropdown would be cleaner for a static form, but here it would add complexity without a functional difference. If the signals forms API adds better primitives for dynamic field sets down the road we can revisit, but for now the current approach handles the dynamic case correctly and stays reactive via the computed.

jadenpuckett (dismissed):

Grumb1e (dismissed):

Grumb1e (approved):

jadenpuckett (approved):

PR #546, opened 2026-04-13
bug[TW29448]: Goal Search - New returns nothing 

dondunne:


plugins/org.eclipse.osee.ats.core/src/org/eclipse/osee/ats/core/query/AbstractAtsQueryImpl.java
```diff
@@ -128,7 +128,37 @@ public <T extends IAtsWorkItem> Collection<T> getItemsNew() {
       if (allArtTypes.contains(AtsArtifactTypes.AbstractWorkflowArtifact)) {
          teamWfs = getTeamWorkflowsNew(allArtTypes, allResults, allArtTypes);
       }
-      return teamWfs;
+
+      /**
+       * These need to be upgraded to use asArtifacts or rolled into the search criteria above
+       */
```
Grumb1e, line 134: It might be worth coming up with a keyword that is easily searchable to find things like this in the future. Something like ATS_TODO or the like.

Grumb1e (approved):

david-w-miller (approved):

PR #560, opened 2026-05-05
bug[TW29451]: Cannot open attached zip files

dondunne:


plugins/org.eclipse.osee.framework.ui.skynet/src/org/eclipse/osee/framework/ui/skynet/render/RenderingUtil.java
```diff
@@ -152,8 +155,13 @@ public static void displayDocument(PresentationType presentationType, Program pr
          if (RenderingUtil.arePopupsAllowed()) {
 
             RenderingUtil.ensureFilenameLimit(contentFile);
-
-            program.execute(contentFile.getLocation().toFile().getAbsolutePath());
+            if (extension.equals("xml")) {
+               // Explicitly tell OS what to open with
+               program.execute(contentFile.getLocation().toFile().getAbsolutePath());
```
Grumb1e, line 153: Will this work regardless of OS and installed programs? Does it have a fallback if what you tell it to open with does not exist?

dondunne: Even though this looks like new code, it was already doing this, so it's a no-change for xml case.  All I'm doing is adding the Program.launch for all other cases since that's what is the best answer for everything else.  It will actually work better for all OS's since it leaves it up to OS.

audrey-denk (dismissed):

jadenpuckett (approved):

Grumb1e (approved):

PR #576, opened 2026-05-18
feature[TW26135]: Implement attribute field for Native Content attribute type 

Grumb1e:


web/apps/osee/src/app/ple/artifact-explorer/lib/components/editor/attributes-editor-panel/attributes-editor-panel.component.ts
```diff
@@ -47,47 +48,89 @@ import { CurrentBranchInfoService } from '@osee/shared/services';
...
+
 	saveChanges() {
 		if (this.updatedAttributes.value.length > 0) {
+			const t = this.tab();
 			const tx: legacyTransaction = {
-				branch: this.tab.branchId,
-				txComment:
-					'Attribute changes for artifact: ' + this.tab.artifact.name,
+				branch: t.branchId,
+				txComment: 'Attribute changes for artifact: ' + t.artifact.name,
```
dondunne: tx comment should be more tied to which operation/editor did the change and not the changed artifact.  what changed is evident in the transaction or change report.  something like "Web Attribute Save".  Also shouldn't be dynamic, if the artifact name is too long you'll exception

web/apps/osee/src/app/shared/components/attributes-editor/native-content-editor/native-content-download.service.ts
```diff
@@ -0,0 +1,51 @@
+/*********************************************************************
+ * Copyright (c) 2025 Boeing
```
dondunne: 2026?

web/apps/osee/src/app/shared/components/attributes-editor/native-content-editor/native-content-editor.component.html
```diff
@@ -0,0 +1,81 @@
+<!--
+/*********************************************************************
+ * Copyright (c) 2025 Boeing
```
dondunne: same

web/apps/osee/src/app/shared/components/attributes-editor/native-content-editor/native-content-editor.component.ts
```diff
@@ -0,0 +1,228 @@
+/*********************************************************************
+ * Copyright (c) 2025 Boeing
```
dondunne: same

web/apps/osee/src/app/shared/dialogs/update-native-content-dialog/update-native-content-dialog.component.html
```diff
@@ -0,0 +1,97 @@
+<!--
+/*********************************************************************
+ * Copyright (c) 2025 Boeing
```
dondunne: same

web/apps/osee/src/app/shared/dialogs/update-native-content-dialog/update-native-content-dialog.component.ts
```diff
@@ -0,0 +1,110 @@
+/*********************************************************************
+ * Copyright (c) 2025 Boeing
```
dondunne: same

web/apps/osee/src/app/ple/artifact-explorer/lib/components/editor/attributes-editor-panel/attributes-editor-panel.component.html
```diff
@@ -45,10 +45,10 @@
 						Applicability:
 					</div>
 					<osee-persisted-applicability-dropdown
```
dondunne, line 47: Kiro code review

----

Here's my review of this PR:

Overall: Solid feature implementation — native content upload/download with a clean dialog, proper separation into a service, and reactive auto-refresh. A few things to tighten up:

Architecture / Design

getArtifactWithRelationsResource calls this.uiService.updateCount() inside the resource factory — this is a side effect inside a reactive computation. The httpResource factory function should be pure (just return a URL or undefined). The updateCount() dependency should be consumed as a signal read, not called imperatively. As written, every time the resource re-evaluates its URL (which happens on any signal change), it increments the update count, potentially causing cascading refreshes.

The .subscribe() calls in native-content-editor.component.ts (both onClickDownload and onClickOpenUpdate) have no error handling. If the HTTP call or file read fails, the user gets no feedback. Add an error callback or pipe through catchError with a snackbar notification.

readFileAsBase64 is called inside switchMap but there's no size validation before reading. The dialog validates size, but if someone bypasses the dialog (or the validation has a race), you could read a huge file into memory as base64. Consider a defensive check.

Code Quality

attributes-editor-panel.component.ts — the panelOpen / togglePanel removal in patch 4 is good cleanup, but the BehaviorSubject<attribute[]> for updatedAttributes is still used imperatively. Consider migrating to a signal for consistency with the rest of the component.

native-content-editor.component.ts line ~155 — the newExt object hardcodes multiplicityId: '3'. This magic string should be a named constant or derived from the attribute type metadata.

update-native-content-dialog.component.ts — BytesPipe is both in imports and providers. Since it's injected via inject(BytesPipe), it only needs to be in providers, not imports.

Minor Issues

Patch 5 removes an //eslint-disable-next-line @typescript-eslint/no-unused-vars from split-applicability.pipe.ts — this is unrelated to the feature. Should be a separate commit or excluded from this PR.

Copyright year: spec files in patch 6 say Copyright (c) 2026 — correct for the current date, just confirming that's intentional vs. a typo of 2025.

CommonModule removal in patch 2 is good — standalone components don't need it. But verify no *ngIf / *ngFor directives were being used from it in the template (looks clean from the template diff).

Testing

The spec files are minimal (just should create). For a feature this complex, I'd expect at least:
Test that onFileSelected rejects oversized files and shows snackbar
Test that submit() returns the selected file
Test that onClickOpenUpdate emits only changed attributes
Test the download service URL construction
Summary: The main concern is item #1 (side effect in the resource factory). The rest are improvements that could be addressed in a follow-up. The overall structure — extracting the download to a service, using signals for pending state, and the auto-refresh pattern — is well thought out.

dondunne (dismissed):

jadenpuckett (dismissed):

jadenpuckett (approved):

david-w-miller (approved):

PR #577, opened 2026-05-19
feature[TW29482]: Update generic report to work with named relation t…

david-w-miller:
…ypes

### DIFF
--- a/plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writers/GenericReportBuilder.java
+++ b/plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writers/GenericReportBuilder.java
@@ -12,6 +12,7 @@
  **********************************************************************/
 package org.eclipse.osee.orcs.rest.internal.writers;
 
+import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import org.eclipse.osee.framework.core.data.ArtifactId;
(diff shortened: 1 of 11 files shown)


plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writers/GenericReportBuilder.java
```diff
@@ -76,6 +77,21 @@ public GenericReport level(String levelName, String typeName) {
       return this;
    }
 
+   @Override
+   public GenericReport relationLevel(String levelName, String relationName, String relationSide) {
+      int depth = 0;
+      if (currentLevel != null) {
+         depth = currentLevel.getDepth() + 1;
+      }
```
Grumb1e: This implementation appears to miss the case where currentLevel is null. In that case, depth would be computed as 0, which conflicts with the expected behavior we discussed.

jadenpuckett:
It might be nice to eventually have this report endpoint have an API that returns a MD document that describes the syntax for writing a generic report. We always have to guess when writing the java code in the report's attribute and some documentation would be nice! This might even encourage users to write their own reports.

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writers/GenericReportBuilder.java
david-w-miller: I had the AI refactor that to make it clearer - instead of code to calculate the next depth the same way every method, it made a private method called nextDepth() that clarifies what is happening e.g. currentLevel.setDepth(nextDepth());

jadenpuckett (approved):

Grumb1e (approved):

PR #581, opened 2026-05-20
feature[TW29503]: Resolve initial web page load slowness.

Grumb1e:


jadenpuckett (approved):
I would suggest sending this to Audrey/Luciano too.

david-w-miller (commented):
What Jules says: Overall Impressions
The PR makes significant structural improvements to UserServiceImpl by moving from eager loading of all users on startup to a lazy, background-loaded approach. This will undoubtedly speed up application startup time and lower the initial resource footprint. Furthermore, refactoring repeated queries into explicit, well-named methods (queryUserById, queryUserByLoginId, etc.) significantly enhances code readability and maintainability.
Key Points for Review
1. Concurrency and Thread Safety
The most significant area of concern lies in how thread state is being managed.
•	The threadToUser Map: The class maintains a ConcurrentHashMap<Thread, UserToken> threadToUser. The PR correctly attempts to address a memory leak here by periodically purging dead threads when the map grows beyond a threshold (if (threadToUser.size() > 100)).
•	Recommendation: A standard java.lang.ThreadLocal<UserToken> is designed exactly for this use case and avoids manual cleanup, memory leaks, and synchronization overhead. Replacing the ConcurrentHashMap with a ThreadLocal will eliminate the need for the custom purge logic in removeUserFromCurrentThread() and simplify setUserForCurrentThread() significantly.
2. Background Loading Logic
•	Guard Condition: The background loader is triggered via triggerBackgroundLoad(). It uses an AtomicBoolean backgroundLoadTriggered to ensure only one thread loads the users.
•	Race Condition in ensureLoaded: In ensureLoaded(), we have:
•	if (!allUsersLoaded) {
•	   for (...) {
•	      cacheUser(toUser(userArtifact));
•	   }
•	   allUsersLoaded = true;
•	}
Since allUsersLoaded is marked volatile, it provides visibility, but the block inside the if is not atomic. If ensureLoaded() is called concurrently from multiple threads (e.g., if a user requests a full list before the background thread finishes), it could lead to multiple queries executing simultaneously or partial cache updates.
•	Recommendation: Use double-checked locking with a synchronized block around the ensureLoaded execution, or defer all background load processing entirely to the daemon thread and ensure direct queries check the DB if the cache misses.
3. String Case Sensitivity and Locales
•	Locale.ROOT: A great improvement is using .toLowerCase(Locale.ROOT) in AuthenticationRequestFilter and UserServiceImpl. This prevents insidious bugs on systems running in locales like Turkish where the uppercase 'I' does not lowercase to 'i'.
4. Cache Operations
•	The cacheUser(UserToken user) method successfully centralizes map updates. It properly checks if (user.isValid() && !user.getIdString().isEmpty()).
•	The cache lookups (like in setUserFromBasic) now appropriately fallback from checking loginIdToUser, then userIdToUser, then accountIdToUser. This is resilient.
5. Exception Handling
•	In toUser(), adding the OseeLog.log when the user artifact fails to convert is an excellent change. It will surface silent failures that previously resulted in a seemingly random SENTINEL token being returned.
Summary
The structural and intent changes are very positive, making the authentication checks more robust and lazily initialized. The primary requested change would be to replace ConcurrentHashMap<Thread, UserToken> with ThreadLocal<UserToken> to natively handle thread-local storage without a custom garbage collection strategy.

Grumb1e:
> What Jules says: Overall Impressions The PR makes significant structural improvements to UserServiceImpl by moving from eager loading of all users on startup to a lazy, background-loaded approach. This will undoubtedly speed up application startup time and lower the initial resource footprint. Furthermore, refactoring repeated queries into explicit, well-named methods (queryUserById, queryUserByLoginId, etc.) significantly enhances code readability and maintainability. Key Points for Review
> 
>     1. Concurrency and Thread Safety
>        The most significant area of concern lies in how thread state is being managed.
>        •	The threadToUser Map: The class maintains a ConcurrentHashMap<Thread, UserToken> threadToUser. The PR correctly attempts to address a memory leak here by periodically purging dead threads when the map grows beyond a threshold (if (threadToUser.size() > 100)).
>        •	Recommendation: A standard java.lang.ThreadLocal is designed exactly for this use case and avoids manual cleanup, memory leaks, and synchronization overhead. Replacing the ConcurrentHashMap with a ThreadLocal will eliminate the need for the custom purge logic in removeUserFromCurrentThread() and simplify setUserForCurrentThread() significantly.
> 
>     2. Background Loading Logic
>        •	Guard Condition: The background loader is triggered via triggerBackgroundLoad(). It uses an AtomicBoolean backgroundLoadTriggered to ensure only one thread loads the users.
>        •	Race Condition in ensureLoaded: In ensureLoaded(), we have:
>        •	if (!allUsersLoaded) {
>        •	   for (...) {
>        •	      cacheUser(toUser(userArtifact));
>        •	   }
>        •	   allUsersLoaded = true;
>        •	}
>        Since allUsersLoaded is marked volatile, it provides visibility, but the block inside the if is not atomic. If ensureLoaded() is called concurrently from multiple threads (e.g., if a user requests a full list before the background thread finishes), it could lead to multiple queries executing simultaneously or partial cache updates.
>        •	Recommendation: Use double-checked locking with a synchronized block around the ensureLoaded execution, or defer all background load processing entirely to the daemon thread and ensure direct queries check the DB if the cache misses.
> 
>     3. String Case Sensitivity and Locales
>        •	Locale.ROOT: A great improvement is using .toLowerCase(Locale.ROOT) in AuthenticationRequestFilter and UserServiceImpl. This prevents insidious bugs on systems running in locales like Turkish where the uppercase 'I' does not lowercase to 'i'.
> 
>     4. Cache Operations
>        •	The cacheUser(UserToken user) method successfully centralizes map updates. It properly checks if (user.isValid() && !user.getIdString().isEmpty()).
>        •	The cache lookups (like in setUserFromBasic) now appropriately fallback from checking loginIdToUser, then userIdToUser, then accountIdToUser. This is resilient.
> 
>     5. Exception Handling
>        •	In toUser(), adding the OseeLog.log when the user artifact fails to convert is an excellent change. It will surface silent failures that previously resulted in a seemingly random SENTINEL token being returned.
>        Summary
>        The structural and intent changes are very positive, making the authentication checks more robust and lazily initialized. The primary requested change would be to replace ConcurrentHashMap<Thread, UserToken> with ThreadLocal to natively handle thread-local storage without a custom garbage collection strategy.

Fun "adversarial" AI moment. Claude says: 

> 
> The review comment is partially valid but the recommendation has a significant flaw.
> 
> What's correct: Yes, ThreadLocal is the textbook solution for per-thread state and would eliminate the manual purge logic.
> 
> Why ThreadLocal isn't a clean win here:
> 
> ThreadLocal has its own well-known memory leak problem in servlet containers. In a thread pool (like Tomcat/Jetty), threads are reused across requests. If removeUserFromCurrentThread() is ever missed (exception path, filter chain short-circuit), a ThreadLocal silently leaks the previous request's user into the next request on that thread — a security bug where request B runs as request A's user.
> 
> With the current ConcurrentHashMap<Thread, UserToken> approach:
> 
> If remove is missed, the next request overwrites the entry via put() — same behavior as ThreadLocal in the happy path.
> The map is externally observable (testable, debuggable).
> The purge logic handles the edge case of threads that die outside the pool (e.g., the background loader thread).
> ThreadLocal would also work fine here — the cleanup semantics are equivalent in practice since both rely on removeUserFromCurrentThread() being called. But it doesn't solve a problem that exists. The current approach isn't leaking in a thread-pool environment because pooled threads stay alive (entries get reused) and the purge handles the rare dead-thread case.
> 
> Bottom line: The suggestion is a valid style preference (ThreadLocal is more idiomatic), but it's not a correctness improvement and doesn't eliminate the need for cleanup. The "avoids memory leaks" claim is misleading — ThreadLocal has the same leak risk in pooled environments if remove() isn't called. I'd push back on this as a "must fix" and treat it as a "could refactor later" if the team prefers ThreadLocal idiom.

dondunne (approved):

PR #587, opened 2026-05-29
bug[TW29549]: Create work-around for search by ATS Id returning false results

dondunne:


plugins/org.eclipse.osee.ats.ide/src/org/eclipse/osee/ats/ide/search/quick/AtsQuickSearchQuery.java
```diff
@@ -0,0 +1,347 @@
+/*********************************************************************
+ * Copyright (c) 2004, 2007 Boeing
```
Grumb1e, line 2: Was this file just moved, or does the year need to be updated?

Grumb1e (approved):

jadenpuckett (approved):

PR #595, opened 2026-06-08
feature[TW29505]: WEB - Markdown edit resizing and collapsible preview pane

Grumb1e:


.github/actions/web-setup/action.yml
dondunne, line 1: Consider less commits.  8 seems like alot.

plugins/org.eclipse.osee.ats.rest/src/org/eclipse/osee/ats/rest/internal/demo/AtsDbConfigMarkdownDemoOp.java
```diff
@@ -0,0 +1,184 @@
...
+         rd.errorf("Could not find System Requirements - Markdown folder on branch %s",
+            DemoBranches.SAW_PL.getName());
+         return;
+      }
+
+      for (String childName : Arrays.asList("Robot API Communication Protocol",
+         "Robot API Safety Requirements", "Robot API Performance Requirements")) {
+         ArtifactToken artifact =
+            changes.createArtifact(parent, CoreArtifactTypes.SystemRequirementMarkdown, childName);
+         String markdown = "## " + childName + " \n\nThe system shall provide " + childName
```
dondunne: Consider String.format.  Tends to read easier than concat.

plugins/org.eclipse.osee.ats.rest/src/org/eclipse/osee/ats/rest/internal/demo/AtsDbConfigMarkdownDemoOp.java
```diff
@@ -0,0 +1,184 @@
...
+         createImageArtifact(rd);
+         createWorkingBranch(rd);
+      } catch (Exception ex) {
+         rd.errorf("Error in markdownDemoInit: %s", Lib.exceptionToString(ex));
+      }
+   }
+
+   private void createSystemRequirements(XResultData rd) {
+      IAtsChangeSet changes =
+         atsApi.createChangeSet("Markdown Demo Init - Create System Requirement Markdown Artifacts", SAW_PL);
```
dondunne: transaction comments are meant to be able to locate where in the code the operation took place.  You don't need to describe the change cause that's in the transaction/change report.  Change to something like "Markdown - create sys reqts".  Same with others in this class.

dondunne (commented):
minor improvements

jadenpuckett (dismissed):

.github/actions/web-setup/action.yml
Grumb1e: It will squash on merge.

web/apps/osee/src/app/shared/components/markdown-editor/markdown-editor.component.ts
```diff
@@ -331,7 +409,120 @@ export class MarkdownEditorComponent {
...
+		wrapper.requestFullscreen().catch(() => {
+			this.uiService.ErrorText =
+				'Fullscreen is not available in this context.';
+		});
+	}
+
+	private initFullscreenListener(): void {
+		fromEvent(document, 'fullscreenchange')
+			.pipe(takeUntilDestroyed(this.destroyRef))
+			.subscribe(() => {
```
jadenpuckett, line 444: I am assuming all of these subscriptions close out when toggling in and out of full screen

plugins/org.eclipse.osee.ats.rest/src/org/eclipse/osee/ats/rest/internal/demo/AtsDbConfigMarkdownDemoOp.java
Grumb1e: It was previously using format, but that triggers spot bugs comments because format expects %n rather than /n. Markdown on the other hand requires /n regardless of platform. I can revert to what it was and add an exclusion.

plugins/org.eclipse.osee.ats.rest/src/org/eclipse/osee/ats/rest/internal/demo/AtsDbConfigMarkdownDemoOp.java
Grumb1e: I'll add a more direct reference to where this cam from in code, but a short summary of changes is still helpful. The transaction/change report is not the only context from which this can be viewed. The summary is especially useful to understand what is happening when inspecting the db directly.

web/apps/osee/src/app/shared/components/markdown-editor/markdown-editor.component.ts
Grumb1e: This is a single long-lived subscription created once at init. It listens for all fullscreen change events (enter and exit) throughout the component's lifetime. takeUntilDestroyed cleans it up when the component is destroyed. It does not re-subscribe on each toggle. 👍

dondunne (approved):

jadenpuckett (approved):

PR #596, opened 2026-06-08
bug[TW29572]: Update software requirements volatility metrics report …

jadenpuckett:
…to use orcs query instead of ats cache version no load relations query

### DIFF
--- a/plugins/org.eclipse.osee.ats.ide/src/org/eclipse/osee/ats/ide/metrics/SoftwareReqVolatilityMetricsBlam.java
+++ b/plugins/org.eclipse.osee.ats.ide/src/org/eclipse/osee/ats/ide/metrics/SoftwareReqVolatilityMetricsBlam.java
@@ -19,6 +19,7 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
+import java.util.logging.Level;
 import java.util.zip.GZIPInputStream;
 import javax.ws.rs.core.Response;
 import org.eclipse.core.runtime.IProgressMonitor;
@@ -81,8 +82,8 @@ public class SoftwareReqVolatilityMetricsBlam extends AbstractBlam {
    }
 
    @Override
-   public void widgetCreated(XWidget xWidget, FormToolkit toolkit, Artifact art,
-      SwtXWidgetRenderer swtXWidgetRenderer , XModifiedListener modListener, boolean isEditable) {
+   public void widgetCreated(XWidget xWidget, FormToolkit toolkit, Artifact art, SwtXWidgetRenderer swtXWidgetRenderer,
+      XModifiedListener modListener, boolean isEditable) {
       if (xWidget.getLabel().equalsIgnoreCase(VERSION)) {
          versions = new ArrayList<>();
          versionWidget = (XHyperlabelVersionSelection) xWidget;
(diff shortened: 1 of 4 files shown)


.kiro/steering/architecture-documentation.md
```diff
@@ -0,0 +1,60 @@
+---
+inclusion: auto
+---
+
+# Architecture Documentation Practice
+
+When making significant changes to a feature area (new patterns, architectural decisions, performance optimizations, API changes, or bug-prone areas), update or create a documentation file in `docs/ai/` and a corresponding steering file in `.kiro/steering/` that references it.
+
+## Shared documentation location
```
Grumb1e: Perhaps the actual content here should reside in docs/ai and this file should just point to it. That way codex and other agents do not need to reference the kiro folder. That could create coupling issues down the line.

plugins/org.eclipse.osee.ats.ide/src/org/eclipse/osee/ats/ide/metrics/SoftwareReqVolatilityMetricsBlam.java
```diff
@@ -110,57 +110,66 @@ public void widgetModified(XWidget widget) {
...
-               }
-               if (bwr != null) {
-                  try {
-                     bwr.close();
-                  } catch (IOException e) {
-                     //do nothing
-                  }
-               }
+            } catch (IOException e) {
+               //do nothing
```
Grumb1e: If something goes wrong, do we really want to silently bury it? It probably doesn't warrant a severe popup, but perhaps a mention in the log.

audrey-denk (dismissed):

audrey-denk (dismissed):

.kiro/steering/conditional/ats.md
```diff
@@ -0,0 +1,6 @@
+---
```
Grumb1e: I think that perhaps these kinds of files should go under their own subfolder in steering. There will likely be quite a few which would make finding the higher level steering difficult. We may eventually want to create /conditional /always folders too.

docs/ai/CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+docs/ai/<subsystem>/<area>/<feature-name>.md
+```
+
+**Rules:**
+- Max depth: 3 levels under `docs/ai/`
+- Use lowercase kebab-case for all folder and file names
+- Folder names should match how developers think about the subsystem/area
+- If a level accumulates 3+ docs, group them into a subfolder
+- Don't pre-create empty folders — let the structure emerge as docs are added
+
```
Grumb1e: You may want to add an instruction to create a matching steering file.

docs/ai/CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
+# Architecture Documentation Practice
+
+When making significant changes to a feature area (new patterns, architectural decisions, performance optimizations, API changes, or bug-prone areas), update or create a documentation file in this `docs/ai/` directory.
```
Grumb1e: It may not be necessary, but might be worth adding a trust, but verify line somewhere. That way the AI doesn't take the docs as the whole story (and possibly make invalid assumptions) and will still verify in the code. That will hopefully also guard against out of date docs.

docs/ai/CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Naming
+
+- **Folders** map to the major plugin or subsystem name
+- **Subfolders** map to logical feature areas within that subsystem
+- **Files** are named after the specific feature: `<feature-name>.md`
+
+## For small changes
+
+Typo fixes, simple renames, one-liner bug fixes — don't create docs for these. Only document when the change introduces knowledge that would save someone (human or AI) time in the future.
```
Grumb1e: Or invalidates existing documentation.

audrey-denk (dismissed):

Grumb1e (dismissed):

.kiro/steering/conditional/ats.md
```diff
@@ -0,0 +1,6 @@
...
+inclusion: fileMatch
+fileMatchPattern: "**/ats/**"
+---
+
+#[[file:docs/ai/ats/metrics/software-req-volatility-metrics.md]]
```
Grumb1e: This pattern may result in reading in a lot of unnecessary context when touching anything with /ats/ in the path. Worth some research to see if chunking different sections of ats into separate steering files with unique file match criteria would be beneficial.

audrey-denk (approved):

Grumb1e (approved):

PR #598, opened 2026-06-09
feature[TW29354]: WEB - Implement data controls banner. 

Grumb1e:


jadenpuckett (dismissed):

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/enums/CoreArtifactTypes.java
```diff
@@ -304,7 +304,8 @@ public interface CoreArtifactTypes {
       .zeroOrOne(EmailEncryptionEnabled)
       .zeroOrOne(EmailCertificateLdapUrl)
       .zeroOrOne(EmailCertificateUploadOverrideLink)
-      .zeroOrOne(EmailCertificateInstructions));
```
dondunne, line 307: Kiro comments

Issues to address:

XSS via [innerHTML] — The banner renders server content with [innerHTML]="bannerContent()". Since the content comes from an artifact attribute that admins can edit, this is likely acceptable. But if non-admin users can ever modify this attribute, it's an XSS vector. Consider using Angular's DomSanitizer to sanitize, or document that this attribute must be admin-only.

Text blocks (""") in CreateSystemBranches.java — The project should confirm Java 15+ target. The other text block in this file (EMAIL_CERTIFICATE_INSTRUCTIONS) already uses this syntax, so this is consistent with existing code. Just noting it.

Missing newline at EOF — Both DatastoreEndpointImpl.java and DatastoreEndpoint.java have \ No newline at end of file. Add a trailing newline.

SiteBannerContent attribute alphabetical ordering — It's inserted between AbridgedEmail and EmailCertificateInstructions in CoreAttributeTypes.java. S comes after E, so it should go further down the file to maintain alphabetical order. Currently it breaks the A...E... sequence.

Banner dismiss resets daily — The dismiss uses new Date().toDateString() which means if the admin updates the banner content, users who dismissed the old banner today won't see the new content until tomorrow. Consider hashing the content and storing that with the dismiss state so changes are shown immediately.

No caching/TTL on the REST call — shareReplay({ bufferSize: 1, refCount: true }) caches the response for the app lifetime. If an admin updates the banner, users won't see changes until they reload the app. This is probably fine for a banner, but worth documenting.

Accessibility — The role="alert" on the banner div is good. The dismiss button has aria-label="Dismiss banner" which is correct. Looks good from an a11y perspective.

Minor nits:

SiteBannerConfig.java could be a record (public record SiteBannerConfig(String content) {}) if the project uses Java 16+, but a POJO is fine for Jackson compatibility.
The test should remain dismissed on same-page navigation navigates to /ple again which triggers a fresh banner API call — this correctly tests that localStorage persists the dismiss. Good.
Verdict: Looks good to merge with the EOF newline fix and the attribute ordering fix. The XSS consideration is worth a comment in the code.

dondunne (approved):

jadenpuckett (approved):

PR #600, opened 2026-06-11
bug[TW29585]: REL - CalculateEndInsertion is running into limit durin…

audrey-denk:
…g zenith import

Change-Id: I1669ffa4b20c4f7944a43f25882cc0ba6a6ffb45


david-w-miller (dismissed):

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
```diff
@@ -758,6 +785,27 @@ private List<Object[]> reOrderRelations(RelationTypeToken relType, ArtifactId ar
 
    }
 
+   /**
+    * Redistributes relation orders into the lower portion of the integer range, reserving the upper portion as runway
```
lucianovaglienti: Why would you redistribute relation orders into the lower portion of the integer range when orders approach Integer.MIN_VALUE?

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TxData.java
```diff
@@ -54,6 +54,13 @@ public static enum TxState {
       COMMIT_FAILED;
    }
    private static final long SPACING = (long) Math.pow(2.0, 18.0);
+   private static final int REORDER_THRESHOLD_MAX = (int) (Integer.MAX_VALUE - SPACING * 2);
+   private static final int REORDER_THRESHOLD_MIN = (int) (Integer.MIN_VALUE + SPACING * 2);
+   /**
+    * When reordering, pack existing relations into the lower 25% of the integer range, reserving 75% as runway for
+    * end-insertions.
+    */
+   private static final double REORDER_TAIL_RESERVE_RATIO = 0.75;
```
lucianovaglienti: Should add one for head as well, and just use that so you don't have to constantly recompute 1 - 0.75

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
```diff
@@ -758,6 +785,27 @@ private List<Object[]> reOrderRelations(RelationTypeToken relType, ArtifactId ar
...
+    * for end-insertions. Used when existing orders approach Integer.MAX_VALUE or Integer.MIN_VALUE.
+    */
+   private List<Object[]> reOrderRelationsCentered(RelationTypeToken relType, ArtifactId artA,
```
lucianovaglienti: Why is this a new function? Don't we already have a reorderRelations?

Wouldn't it just be smarter to have a check inside reOrderRelations to re-center if need be?

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
```diff
@@ -758,6 +785,27 @@ private List<Object[]> reOrderRelations(RelationTypeToken relType, ArtifactId ar
...
+      TreeMap<Integer, Pair<ArtifactId, GammaId>> relOrders) {
+
+      long totalRange = (long) Integer.MAX_VALUE - (long) Integer.MIN_VALUE;
```
lucianovaglienti: Trying to understand the math here:

usable range = (MAX_INT - MIN_INT)*(0.25)
padding = ((MAX_INT - MIN_INT)*(0.25))/(# of rel orders + 1)
rel order = MIN_INT + pad

As # of rel orders increase, you would insert earlier and earlier into the integer range(i.e. not insert at end?)
Can you confirm my understanding here that we are just essentially doing a sliding window?

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
```diff
@@ -723,6 +735,21 @@ private int reOrderRelTypeArtA(RelationTypeToken relType, ArtifactId artA) {
       return resultsCount;
    }
 
+   /**
+    * Reorders all relations for a given relType/artA by centering them across the full integer range. This is used when
+    * existing order values have drifted too close to Integer.MAX_VALUE or Integer.MIN_VALUE, leaving no room for new
+    * insertions.
+    */
+   private int reOrderRelTypeArtACentered(RelationTypeToken relType, ArtifactId artA) {
```
lucianovaglienti: Why do we have a new function here?

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
```diff
@@ -723,6 +735,21 @@ private int reOrderRelTypeArtA(RelationTypeToken relType, ArtifactId artA) {
...
+
+      List<Object[]> reOrderData = reOrderRelationsCentered(relType, artA, getRelationSideAData(relType, artA));
+
+      int resultsCount = orcsApi.getJdbcService().getClient().runBatchUpdate(
```
lucianovaglienti: I would think you would want whatever batch update that needs to occur with the rest of the relation update in TxBuilder

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
```diff
@@ -638,6 +638,12 @@ private void relate(RelationTypeToken relType, ArtifactId artA, ArtifactId artB,
             txData.addRelationSideA(relType, artA, relOrders);
          }
          if (insertType.equals("start")) {
+            if (minOrder < TxData.getReorderThresholdMin()) {
```
lucianovaglienti: I would think ReorderThresholdMin, and ReorderThresholdMax needs to be non-static values or else you might get caught in a loop here, i.e. the max and min shift based on your relation list length

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
audrey-denk: I think the wording here is bad.  "lower" refers to the higher numbers.  If you think of the order at top to bottom... higher numbers are less than lower numbers.  yeah... I'll reword that.

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
audrey-denk: yes you would insert earlier and earlier only because you are trying to leave more room for adding to the end of the list.  The hope here is that you do this very infrequently because you are leaving room at the end so that we don't brush up against the limit any time soon.

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
audrey-denk: Not sure what you mean here.  This batch update is an "update" statement where afterwards it will be an insert.

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
audrey-denk: done.

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TxData.java
audrey-denk: Done.

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
lucianovaglienti: When commit occurs, the transaction builder does a larger batch update as part of the transaction. You could delay the batch update here by storing it, and having that section of the transaction builder update it.

audrey-denk: my understanding of batch updates is they are tied to a particular sql statement.  The "update" batch statement updates all of the existing rel_orders for the relation; where the insert that results from the new relation is a different commit altogether.  I don't think they can be combined.

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TransactionBuilderImpl.java
audrey-denk: Yeah.. this could be a problem.... making it dynamic.

plugins/org.eclipse.osee.ats.ide.integration.tests/src/org/eclipse/osee/ats/ide/integration/tests/orcs/rest/RelationOrderBoundaryTest.java
```diff
@@ -0,0 +1,238 @@
...
+    * <p>
+    * Strategy:
+    * <ol>
+    * <li>Create a side-A artifact with a few initial side-B relations</li>
+    * <li>Directly update the rel_order values in the DB to near Integer.MAX_VALUE</li>
+    * <li>Add more relations — this should trigger the reorder logic</li>
+    * <li>Verify all relations exist and orders are valid (no duplicates, all ascending)</li>
+    * </ol>
+    */
+   @Test
```
lucianovaglienti, line 61: A nit for all these tests: should probably create a branch to perform them on so they can be executed independently.

plugins/org.eclipse.osee.orcs.core/src/org/eclipse/osee/orcs/core/internal/transaction/TxData.java
```diff
@@ -231,7 +233,40 @@ public int calculateEndInsertionOrderIndex(int currentEndIndex) {
...
+      return (int) (((long) afterIndex + (long) beforeIndex) / 2);
+   }
+
+   /**
+    * Returns true if a reorder is needed — either less than SPACING room below currentMin (head insertions) or less
+    * than SPACING room above currentMax (end appends).
+    */
+   public static boolean shouldReorder(int currentMin, int currentMax) {
+      boolean lowOnHeadRoom = (long) currentMin - (long) Integer.MIN_VALUE < SPACING;
+      boolean lowOnTailRoom = (long) Integer.MAX_VALUE - (long) currentMax < SPACING;
```
lucianovaglienti, line 246: Just a thought...might want to look out for math under/over flow

lucianovaglienti (dismissed):

lucianovaglienti (approved):

plugins/org.eclipse.osee.ats.ide.integration.tests/src/org/eclipse/osee/ats/ide/integration/tests/orcs/rest/RelationOrderBoundaryTest.java
```diff
@@ -0,0 +1,243 @@
+/*********************************************************************
+ * Copyright (c) 2024 Boeing
```
dondunne, line 2: 2026

david-w-miller (approved):

dondunne (approved):

PR #602, opened 2026-06-12
Add cold storage

audrey-denk:


plugins/org.eclipse.osee.orcs.rest.model/src/org/eclipse/osee/orcs/rest/model/TransactionEndpoint.java
```diff
@@ -179,4 +179,93 @@ List<ChangeItem> getArtifactHistory(@PathParam("art-id") ArtifactId artifact,
...
+    * osee_tx_details, and osee_branch rows to a compressed binary file, then purges them from the database.
+    *
+    * @param limit maximum number of branches to archive per invocation (default 100)
+    * @param retentionDays minimum age in days since last transaction before a branch is eligible (default 365)
+    */
+   @POST
+   @Path("cold/archive")
+   @Produces(MediaType.APPLICATION_JSON)
+   XResultData archiveToColdStorage(@QueryParam("limit") @DefaultValue("100") int limit,
+      @QueryParam("retentionDays") @DefaultValue("365") int retentionDays);
```
jadenpuckett, line 196: I would like each of these endpoints to require admin role. I don't see this anywhere.

jadenpuckett (changes requested):

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/TxPurgeColdStorage.java
```diff
@@ -0,0 +1,775 @@
...
+
+         if (!found) {
+            results.errorf("Transaction %s not found in archive file %s", txId, fileName);
+         } else {
+            results.logf("Successfully restored transaction %s from %s", txId, fileName);
+         }
+
+      } catch (IOException ex) {
+         results.errorf("Failed to read cold storage file: %s", ex.getMessage());
+         throw OseeCoreException.wrap(ex);
```
david-w-miller: What is the goal here - should the rest user be looking for thrown exceptions or should they be checking the results for exceptions? If there is an immediate re-throw, then the results will not be returned.

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/TxsColdStorage.java
```diff
@@ -0,0 +1,862 @@
...
+
+         // Insert catalog row (branch will be deleted by purgeBranch after this returns)
+         jdbcClient.runPreparedUpdate(INSERT_CATALOG, branchId.getId(), branchName[0], fileName, totalTxsRows,
+            totalTxDetailsRows, branchState[0]);
+
+         results.logf("Archived branch %d (%s) to cold storage: %s (%d bytes, %d txs rows, %d tx_details rows)",
+            branchId.getId(), branchName[0], fileName, fileSize, totalTxsRows, totalTxDetailsRows);
+
+      } catch (IOException ex) {
+         results.errorf("Failed to write cold storage file: %s", ex.getMessage());
```
david-w-miller, line 363: Similar to TxsPurgeColdStorage.java, if you throw an exception, then the user won't see the results, so it should either be results.errorf or throw, but not both.

jadenpuckett (approved):

david-w-miller (approved):
Looks good

PR #614, opened 2026-06-30
feature[TW29697]: Update artifact explorer to use gamma ids

jadenpuckett:


web/apps/osee/src/styles.sass
```diff
@@ -27,6 +27,33 @@ body
...
+osee-artifact-explorer
+  --mdc-filled-text-field-container-color: transparent
+  --mat-form-field-container-color: transparent
+  --mdc-filled-text-field-hover-container-color: transparent
+  --mdc-filled-text-field-disabled-container-color: transparent
+  .mdc-text-field--filled
+    background: transparent !important
+  .mdc-text-field--filled::before
+    display: none !important
+
```
Grumb1e, line 44: Is there another way to accomplish the styling you want with public api rather than internal material classes that could change during an angular update?

---- AI

Fragile (risky across Angular Material updates):

Selector / Pattern	Risk	Why
.mat-mdc-menu-item	Medium	Internal class name. Not a public API — could be renamed or restructured.
.mat-icon	Low-Medium	Relatively stable but still an internal class, not a CSS token.
.mdc-form-field	High	MDC internal class. Angular Material wraps MDC Web and can change how these are rendered.
.mdc-form-field > label	High	Depends on exact DOM structure (direct child). If Angular wraps the label in an extra element, this breaks.
.mdc-text-field--filled	High	Internal MDC class targeting a specific variant's DOM.
.mdc-text-field--filled::before	High	Relies on a specific pseudo-element implementation detail (the ripple/background layer). If Angular Material changes how the filled background is rendered, this silently stops working.

web/apps/osee/src/app/ple/artifact-explorer/lib/components/hierarchy/artifact-hierarchy-panel/artifact-hierarchy-panel.component.ts
```diff
@@ -12,46 +12,70 @@
...
+				);
+				--mat-minimal-pseudo-checkbox-selected-checkmark-color: var(
+					--osee-primary-default
+				);
+			}
+			.hierarchy-scroll::-webkit-scrollbar {
+				width: 12px;
+				height: 12px;
+			}
+			.hierarchy-scroll::-webkit-scrollbar-thumb {
```
Grumb1e, line 62: --- AI ---
The only caveat on the scrollbar styling is that ::-webkit-scrollbar is non-standard and Chromium-only. The CSS Scrollbars spec (scrollbar-width, scrollbar-color) is the standards-track approach. But that's a browser evolution concern

Grumb1e (commented):
Most of the below is noise, but some of it is valuable. You can paste the review into your terminal and have it determine what is valid and execute fixes. 

--- AI Review ---

Code Review
This is a substantial refactor of the Artifact Explorer with a new VS Code-like layout (activity bar + resizable panels), new artifact history feature, auto-saving attribute editor, dirty state tracking, and supporting backend changes. Here are my findings grouped by file:

Java Backend
ArtifactEndpointImpl.java
error — Removed getPathToArtifact recursive CTE query, replaced with parent-walking loop

The new implementation walks up the hierarchy one artifact at a time (N+1 queries for N-deep trees, up to 50 levels). The old implementation used a single recursive CTE query. This is a significant performance regression for deeply nested hierarchies, as each asArtifactOrSentinel() call is a full database query.

while (maxDepth-- > 0) {
   path.add(currentId);
   ArtifactReadable current =
      orcsApi.getQueryFactory().fromBranch(branch, viewId).andId(currentId).asArtifactOrSentinel();
Suggestion: Consider restoring the CTE-based approach or using a bulk query to walk the hierarchy rather than individual lookups.

warning — getArtifactHistoryPaginated loads ALL history then paginates in-memory

List<ChangeItem> history = orcsApi.getTransactionFactory().getArtifactHistory(artifact, branch);
This loads the entire artifact history into memory before sorting, grouping, and slicing. For artifacts with thousands of transactions, this is an O(n) memory and time operation for every page request.

Suggestion: Consider adding a paginated query at the TransactionFactory level, or at minimum cache the full history with a TTL.

warning — restoreArtifact uses replaceWithBaselineTxVersion without validation

The endpoint accepts a txId string and calls Long.parseLong(txId) without validating that the transaction belongs to the specified branch or artifact. A malformed or out-of-scope transaction ID could produce unexpected behavior.

Suggestion: Add validation that the transaction ID exists and is associated with the given artifact/branch before restoring.

ArtifactWithRelations.java
suggestion — Reflection-based value extraction in toStringPojo

Object nameValue = value.getClass().getMethod("getName").invoke(value);
Using reflection here works but is fragile and has performance implications. Consider using an interface check (if (value instanceof Named)) or the framework's existing type hierarchy.

ArtifactHistoryResult.java
suggestion — Clean implementation. No issues.

Angular — Artifact Explorer Component
artifact-explorer.component.ts
warning — onDividerMouseDown does not account for the activity bar width

The percentage calculation uses the full container width, but the activity bar (tw-w-12, 48px) is part of the container. This means the panel width calculation is slightly off. Not a blocker, but can produce a small jump on first drag.

suggestion — panelWidthPercent and panelCollapsed are not persisted

Users who resize/collapse the panel will lose their setting on page reload. Consider saving to localStorage or the user preferences service.

artifact-explorer.component.html
suggestion — role="separator" on the resize divider is good for accessibility, but aria-valuenow is bound to the signal — ensure it's the correct ARIA role for a separator (it should have aria-orientation="vertical" which is correctly set).

Angular — Artifact Tab Service
artifact-explorer-tab.service.ts
error — removeTab index correction can set index to -1

removeTab(index: number) {
    this.tabs.update((rows) => rows.filter((_, i) => index !== i));
    if (this._selectedIndex() >= this.tabs().length) {
        this._selectedIndex.set(Math.max(0, this.tabs().length - 1));
    }
}
If all tabs are removed (length becomes 0), Math.max(0, -1) yields 0, but there are 0 tabs. The selectedIndex should be set to -1 or handled specially when no tabs exist. Currently the empty state is handled by the template (tabs().length === 0), so this is unlikely to crash, but it's semantically wrong.

bug — addTab findIndex uses branchId variable that is the input parameter

const currentIndex = this.tabs().findIndex(
    (existingTab) =>
        existingTab.tabType === 'Artifact' &&
        existingTab.branchId === branchId &&
        existingTab.artifact?.id === artifact.id
);
This is actually a fix (previously it used this.uiService.id.value which could be stale). Good change.

Angular — Attributes Editor Panel
attributes-editor-panel.component.ts
warning — _nameSyncEffect has a side effect inside a computed-like effect

private _nameSyncEffect = effect(() => {
    const name = this.artifactResource.value()?.name;
    if (name && name !== this.tab().artifact.name) {
        this.tabService.updateTabTitle(this.tab().artifact.id, name);
    }
});
This calls this.tab() inside the effect, which reads from a signal and creates a dependency on tab. If the tab input changes, this effect re-fires. This is likely intended but should be documented. Also, tab().artifact.name won't update (it's an input snapshot), so this could fire repeatedly.

warning — _lastAttributes used as a side-effect cache inside computed

protected attributes = computed<attribute<string, ATTRIBUTETYPEID>[]>(() => {
    const resourceAttrs = this.artifactResource.value()?.attributes;
    if (resourceAttrs) {
        this._lastAttributes = [...resourceAttrs].sort(/*...*/);
    }
    return this._lastAttributes ?? [...this.tab().artifact.attributes].sort(/*...*/);
});
Mutating an external variable (_lastAttributes) inside a computed violates Angular signals best practices. computed functions should be pure. This works today but can produce confusing bugs if the signal graph changes. Consider using linkedSignal or a separate effect to track the last-known-good value.

persisted-artifact-attribute-editor.component.ts
suggestion — The effect that tracks previousValue depends on displayValue() and saving()

constructor() {
    effect(() => {
        if (!this.saving()) {
            this.previousValue.set(this.displayValue());
        }
    });
}
When the resource refreshes (e.g., after another tab changes the same attribute), this will update previousValue, which is correct. But if saving is true and the resource updates, the new value is missed. The saving flag is reset in the subscribe handler, which should trigger the effect again. This is sound but subtle — worth a comment in the code.

suggestion — No debounce on onValueChange for FocusLostInput

If FocusLostInputComponent emits on every keystroke rather than on blur only, rapid saves could fire. Verify that FocusLostInput only emits on blur/focus-lost. If it fires on each change, add debounce.

Angular — Artifact History Panel
artifact-history-panel.component.ts
warning — historyResource is created in the class body

protected historyResource = this.httpService.getArtifactHistoryResource(
    this.branchId, this.artifactId, this.currentPage, this.pageSize
);
This httpResource is created once during construction. If tab() input changes (e.g., user clicks a different tab), branchId and artifactId are computed from tab(), so the resource should automatically re-fetch. This is correct.

artifact-history-diff-dialog.component.ts
warning — confirm() is used for the revert confirmation

if (!confirm(message)) return;
Using window.confirm() in an Angular application breaks testability and is inconsistent with the rest of the UI which uses MatDialog. Consider a confirmation dialog component.

warning — No error handling on modifyArtifactAndMutate failure in revert

The subscribe has no error callback. If the revert transaction fails, the dialog closes silently (it doesn't close, actually — it only closes on 'restored'). But there's no user feedback on failure either.

Angular — Artifact Hierarchy Component
artifact-hierarchy.component.ts
suggestion — Removed repeat({ delay: () => this.uiService.update }) pattern, replaced with startWith(true) + debounceTime(100)

This is a reasonable approach for refresh-on-update, but the debounce means rapid UI updates might coalesce. This is likely desired.

warning — latestPaths no longer auto-expands children along the path

The old code had:

childPaths.forEach((path) => {
    arts.find((art) => {
        if (art.id === path[path.length - 1]) {
            this.expandArtifact(art.id);
        }
    });
});
This was moved to ArtifactHierarchyPathService.expandAlongPaths(). Verify that the expansion still triggers for all hierarchy levels (the new approach expands from the path service, not from within each hierarchy component instance).

Angular — Tab Group Component
artifact-tab-group.component.ts
warning — removeTab dirty check is too broad

removeTab(index: number) {
    if (this.dirtyService.hasDirtyEditors()) {
        const confirmed = confirm('You have unsaved changes...');
This checks if any editor is dirty, not specifically the tab being closed. If you have 3 tabs open and tab 1 is dirty, closing tab 3 (which is clean) still triggers the warning.

Suggestion: Check if the specific tab being closed has dirty state.

Angular — Expanded Service
artifact-hierarchy-artifacts-expanded.service.ts
error — Two effects call this.artifactsExpandedStructArray.set([]) which creates a potential circular dependency

private _branchClearEffect = effect(() => {
    this._branchAndView();
    this.artifactsExpandedStructArray.set([]);
});

private _viewClearEffect = effect(() => {
    this._viewReset();
    this.artifactsExpandedStructArray.set([]);
});
Both effects write to the same signal. Angular effects that write to signals can trigger unexpected re-runs. These are fine individually, but if branch and view change simultaneously (common when switching branches), both effects fire and both clear the array. Not a bug per se, but effect writing to signals is flagged by Angular's linter. Use linkedSignal or an explicit subscription pattern instead.

Playwright E2E Tests
markdown-editor.e2e-spec.ts
suggestion — Comment removal throughout

Many comments were stripped. While the code is fairly readable without them, some test intent comments (like // Should NOT contain the text outside the table) aided quick scanning. Consider keeping intent comments on non-obvious assertions.

suggestion — Several tests now use .first() and .last() on button collections without assertions

await dialog.locator('button:not([disabled])').filter({ hasText: 'merge_type' }).first().click();
If the DOM structure changes and no button matches, .first() will fail with a generic timeout error. Adding explicit await expect(...).toBeVisible() before clicking would make failures more diagnosable.

helpers.ts
suggestion — createBranchViaApi has hardcoded SAW_BLD_1_BRANCH_ID = '3'

This is fine for test setup but fragile if the test database changes. Document this assumption.

suggestion — searchAndOpenArtifact uses page.locator('osee-artifact-search button') which is component-structure-dependent

If the component template changes, this locator breaks. Consider adding a data-testid attribute to search results.

Theme / SCSS
m3-theme.scss
suggestion — Dark mode error palette repeats values at slots 60/70 and 70/80

60: var(--osee-red-7),
70: var(--osee-red-8),
80: var(--osee-red-8),
Slot 60 and 70 both use --osee-red-7 and --osee-red-8 respectively, and slots 70/80 overlap. This flattens the contrast ramp in that range. Verify this is intentional for the desired vivid-red appearance.

General Observations
Type migration from attribute to attribute<string, ATTRIBUTETYPEID> is consistently applied across the codebase. The deprecated re-export in 
attribute.ts
 provides backward compatibility.

Removal of ExpansionPanelComponent in favor of direct visibility toggling ([class.tw-hidden]) is a good UX improvement for the editor sections — it preserves component state.

Removal of ArtifactWithRelationsAttribute Java class in favor of AttributePojo<String> simplifies the model and adds gammaId which is needed for the new relation operations.

The panelLocation user preference removal is a breaking change for existing users who had left-panel configured. The settings dialog now shows "No configurable settings" which is fine, but the migration path silently loses the old preference without notification.

No barrel import violations detected. The new imports from @osee/attributes/types and @osee/attributes/constants follow the project's module structure.

Summary of Severity
Severity	Count	Key Items
Error	3	N+1 query in path-to-artifact, effects writing signals, removeTab index off-by-one
Warning	9	In-memory pagination, dirty check too broad, confirm() usage, impure computed, no error feedback
Suggestion	11	Persistence, debounce, test robustness, comment preservation, theme slots

docs/ai/web/web-coding-standards.md
```diff
@@ -33,6 +33,37 @@ Generate code for our repo using **Angular (v18+)**. **Follow our repo style exa
 - For dialogs, return values via `MatDialogRef.close(value)`.
 - Use explicit interfaces for data passed via `MAT_DIALOG_DATA`.
```
Grumb1e, line 34: Might be worth a mention to use the new colors/stylings you added in this PR where reasonable to improve consistency.

web/apps/osee/src/app/ple/artifact-explorer/lib/components/hierarchy/artifact-hierarchy-panel/artifact-hierarchy-panel.component.ts
jadenpuckett: Added a standard default for Firefox/Safari. Chrome/Edge can use the webkit spec still.

web/apps/osee/src/styles.sass
jadenpuckett: The CSS custom property tokens (--mdc-filled-text-field-container-color: transparent) are set on the component host and in the global styles, but they don't cascade through Angular's view encapsulation into child components' internal shadow-like boundaries. The Material form field reads its container color from a variable that's resolved at the internal template level, not inherited from ancestor elements.

We tried:

:host { --mdc-filled-text-field-container-color: transparent } on the parent component — doesn't cascade into child component templates
Inline style="--mdc-filled-text-field-container-color: transparent" on the <mat-form-field> element — Angular Material doesn't read tokens from inline styles on the host element
appearance="outline" — works but changes the visual design (removes underline, adds border)
The only working approach is the direct class override in global styles, scoped by the osee-artifact-explorer ancestor selector so it only affects form fields within the artifact explorer — not app-wide. This is a known limitation of Angular Material's current theming system where CSS custom properties don't pierce component boundaries reliably for the filled text field variant.

If a future Angular Material version fixes token inheritance, these direct overrides can be removed and the tokens alone will suffice (they're already declared above the fallback).

We added a code comment to the styles.sass to update this in a future Angular Material version.

docs/ai/web/web-coding-standards.md
jadenpuckett: done

Grumb1e (approved):

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/data/AttributeTypeInputStream.java
```diff
@@ -47,12 +49,11 @@ public InputStream valueFromStorageString(String storedValue) {
 
    @Override
    public String storageStringFromValue(InputStream value) {
-      throw new UnsupportedOperationException();
+      return "";
```
david-w-miller, line 52: The added comment says it should not be read from, the exception looked to enforce that, why not keep it?

david-w-miller (approved):
Very minor comment, everything looks good to me.

PR #617, opened 2026-07-06
bug[TW29653]: REL - add fx so that rel_order and top_rel_order are al…

audrey-denk:
…ways present

Change-Id: I2a5e0a34e5de6652839969c5628702601b148558

### DIFF
--- a/plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/search/handlers/RelatedToSqlHandler.java
+++ b/plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/search/handlers/RelatedToSqlHandler.java
@@ -165,8 +165,13 @@ public class RelatedToSqlHandler extends SqlHandler<CriteriaRelatedTo> {
 
    @Override
    public void writeSelectFields(AbstractSqlWriter writer) {
-      if (this.criteria.getRelationTypeSide().isNewRelationTable()) {
-         writer.write(", rel_type as top_rel_type, rel_order as top_rel_order");
+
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/search/handlers/RelatedToSqlHandler.java
```diff
@@ -165,8 +165,13 @@ public int getPriority() {
 
    @Override
    public void writeSelectFields(AbstractSqlWriter writer) {
-      if (this.criteria.getRelationTypeSide().isNewRelationTable()) {
-         writer.write(", rel_type as top_rel_type, rel_order as top_rel_order");
+
+      if (writer.getFirstAlias(OseeDb.RELATION_TABLE).equals(relAlias)) {
```
lucianovaglienti, line 169: Just a question, but might we want to see if there is any alias that equals relAlias (i.e. instead of just the first one)

lucianovaglienti (dismissed):

plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/search/handlers/RelatedToSqlHandler.java
audrey-denk: yes, but I need at least one.  So my thought here was that I only want to add if we are on the first one.  After that, don't cause that would cause  0 rel_type, 0 top_rel_order, 0 rel_type, 0 top_rel_order etc etc

jadenpuckett (dismissed):

lucianovaglienti (approved):

jadenpuckett (approved):

PR #628, opened 2026-07-22
feature[TW29707]: Create Reload-All for Workflow Editor

dondunne:


plugins/org.eclipse.osee.ats.api/src/org/eclipse/osee/ats/api/util/AtsImage.java
Grumb1e: Does not appear to be reassigned to another list, so should be final.

Grumb1e (commented):
You should take a quick look at the spotbugs comments and resolve or filter them.

plugins/org.eclipse.osee.ats.api/src/org/eclipse/osee/ats/api/util/AtsImage.java
dondunne: done

Grumb1e (approved):

audrey-denk (approved):

PR #631, opened 2026-07-23
bug[TW29384]: Workflow Editor title banner and twisties are not colored correctly

dondunne:


plugins/org.eclipse.osee.framework.ui.plugin/src/org/eclipse/osee/framework/ui/plugin/util/AWorkbench.java
```diff
@@ -48,92 +48,29 @@
...
+      boolean darkMode = false;
+      try {
+         darkMode = Display.isSystemDarkTheme();
+         if (!darkMode) {
+            try {
+               Object engine = Display.getDefault().getData("org.eclipse.e4.ui.css.swt.theme");
+               if (engine != null) {
+                  // Use reflection to call engine.getActiveTheme().getId()
+                  Object activeTheme = engine.getClass().getMethod("getActiveTheme").invoke(engine);
+                  if (activeTheme != null) {
```
audrey-denk, line 61: Just the general case where can we return a sentinel or something so we don't have to compare with null?

plugins/org.eclipse.osee.ats.ide/src/org/eclipse/osee/ats/ide/editor/tab/workflow/section/WfeWorkflowSection.java
```diff
@@ -148,6 +149,9 @@ private void refreshStateTitle() {
       String currentStateTitle =
          getCurrentStateTitle(sma, statePage.getName(), isCurrentState(), statePage.isCancelled());
       section.setText(currentStateTitle);
+      section.setTitleBarBackground(new Color(section.getDisplay(), 220, 235, 252));
+      section.setTitleBarBorderColor(null);
+      section.setTitleBarGradientBackground(null);
       if (sma.isInState(statePage)) {
          section.setTitleBarForeground(Displays.getSystemColor(SWT.COLOR_DARK_GREEN));
       } else {
```
Grumb1e, line 157: Rather than explicitly setting colors like this, shouldn't this be using  CSS stylesheet and targeting widgets by their data-binding class or ID with theme attributes?

/* 1. Default Styles (Light Mode) */
.MyCustomWidget {
    background-color: #FFFFFF;
    color: #1A1A1A;
}

/* 2. Overrides (Dark Mode Theme ID) */
[theme_id="org.eclipse.e4.ui.css.theme.e4_dark"] .MyCustomWidget {
    background-color: #262626;
    color: #F0F0F0;
}

audrey-denk (approved):
Just left a comment about comparing with null which we try to avoid.

Grumb1e (approved):
If we plan to keep the client around, I still think we should move to using CSS stylesheets so we don't keep adding technical debt.  I would imagine kiro can make that happen without too much trouble. Since you manage the client UI, I will leave that call up to you though.👍

PR #635, opened 2026-07-29
feature[TW29610]: Add sort capability to generic report

david-w-miller:
This adds two new capabilities - a sort method for the generic report, and a REST endpoint to put some sortable numbers matching the hierarchy depth into a given attribute.

### DIFF
--- a/plugins/org.eclipse.osee.orcs.rest.model/src/org/eclipse/osee/orcs/rest/model/ReportEndpoint.java
+++ b/plugins/org.eclipse.osee.orcs.rest.model/src/org/eclipse/osee/orcs/rest/model/ReportEndpoint.java
@@ -15,15 +15,18 @@ package org.eclipse.osee.orcs.rest.model;
 
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
+import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
 
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import org.eclipse.osee.framework.core.data.ArtifactId;
 import org.eclipse.osee.framework.core.data.BranchId;
 import org.eclipse.osee.framework.jdk.core.annotation.Swagger;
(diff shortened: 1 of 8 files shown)


plugins/org.eclipse.osee.orcs.rest.model/src/org/eclipse/osee/orcs/rest/model/ReportEndpoint.java
```diff
@@ -47,4 +49,11 @@ Response getReportFromTemplateAsync(@PathParam("branch") BranchId branch,
       @DefaultValue("-1") @PathParam("template") ArtifactId templateArt,
       @PathParam("email") String emailRecipient);
 
+   @PUT
+   @Path("{branch}/hierarchyNumber/{artifact}/attributeType/{attributeType}/padding/{padding}")
```
Grumb1e: Should padding be a query param with a sensible default? It seems strange for that to be a path param that must be supplied; especially when there is an arbitrary maximum of 5.

jadenpuckett (dismissed):

jadenpuckett (dismissed):

plugins/org.eclipse.osee.orcs.rest.model/src/org/eclipse/osee/orcs/rest/model/ReportEndpoint.java
```diff
@@ -47,4 +50,11 @@ Response getReportFromTemplateAsync(@PathParam("branch") BranchId branch,
       @DefaultValue("-1") @PathParam("template") ArtifactId templateArt,
       @PathParam("email") String emailRecipient);
 
+   @PUT
+   @Path("{branch}/hierarchyNumber/{artifact}")
+   @Produces({MediaType.APPLICATION_JSON})
+   XResultData applyHierarchyNumbers(@PathParam("branch") BranchId branch,
+      @PathParam("artifact") ArtifactId startArtifact, @QueryParam("attributeType") long attributeTypeId,
```
Grumb1e, line 57: A summary of thoughts from our discussion:

- Limit to admin only. 
- Limit possible accepted attribute types. 
- Early check that attribute type has been provided and is one of the approved options.

jadenpuckett (approved):

Grumb1e (approved):

PR #640, opened 2026-08-10
feature[TW29610]: Add html and xlsx output to Generic Report

david-w-miller:


plugins/org.eclipse.osee.framework.core/OSEE-INF/markdown/markdownToExcelStyles.css
jadenpuckett, line 1: I would recommend consolidating the .css for all markdown exports/publishes. We already have a .css somewhere else for flexmark. We need 1 source of truth.

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/publishing/markdown/MarkdownTableToPdfImage.java
jadenpuckett, line 1: Rename this file to something like MarkdownTableToPdfImage or something similar to be more explicit

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/publishing/markdown/MarkdownTableRenderer.java
```diff
@@ -0,0 +1,185 @@
...
+   private String getCssStyles() {
+      try {
+         return OseeInf.getResourceContents("markdown/" + CSS_RESOURCE + ".css", MarkdownTableRenderer.class);
+      } catch (Exception ex) {
+         // Fallback to basic table styling if resource not found
+         return getDefaultTableCss();
+      }
+   }
+
+   private static String getDefaultTableCss() {
```
jadenpuckett: css needs to live somewhere else

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writers/PublishTemplateReportHtml.java
```diff
@@ -0,0 +1,310 @@
...
+      writer.write("</h1>\n<ul>\n");
+      for (String result : results.getResults()) {
+         writer.write("<li>");
+         writer.write(escapeHtml(result));
+         writer.write("</li>\n");
+      }
+      writer.write("</ul>\n</body>\n</html>\n");
+   }
+
+   private void writeStyles(Writer writer) throws IOException {
```
jadenpuckett, line 261: It might be easier to have large html chunks in html files in the same package/bundle somewhere else? For readability?

jadenpuckett: At lease enumerations and resources make things more readable and organized

jadenpuckett (changes requested):

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/publishing/markdown/MarkdownHtmlUtil.java
```diff
@@ -229,4 +239,27 @@ private static StringModificationResult processCaptions(String content, String c
...
 
+   /**
+    * Determines whether a string contains markdown content that benefits from rich rendering (tables, images, or
+    * formatted text). Shared heuristic used by both Excel and HTML report writers.
+    *
+    * @param value the cell/content value to check
+    * @return true if the value contains renderable markdown constructs
+    */
+   public static boolean containsMarkdownContent(String value) {
+      if (value == null || value.isEmpty()) {
```
dondunne: Consider if (Strings.isInvalid(value)) { return false; }

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/publishing/markdown/MarkdownHtmlUtil.java
```diff
@@ -229,4 +239,27 @@ private static StringModificationResult processCaptions(String content, String c
       return new StringModificationResult(sb.toString(), changes);
    }
 
+   /**
+    * Determines whether a string contains markdown content that benefits from rich rendering (tables, images, or
+    * formatted text). Shared heuristic used by both Excel and HTML report writers.
+    *
```
dondunne: Consider removing extraneous java doc.  value and return should be obvious by method signature.  Review other methods also.

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/publishing/markdown/MarkdownTableToPdfImage.java
```diff
@@ -0,0 +1,185 @@
...
+         return pdfToImage(pdfBytes);
+      } catch (Exception ex) {
+         throw new OseeCoreException("Failed to render markdown to image", ex);
+      }
+   }
+
+   /**
+    * Renders an HTML fragment to a PNG image. Useful when the caller has already converted markdown to HTML and wants to
+    * render a specific block (like a table) as an image.
+    *
```
dondunne, line 75: good example 1) htmlFragment javadoc not needed 2) use Png in method name and remove javadoc  renderHtmlToPngImage 3) Don't need to ever know what it throws, IDE handles all the for you.  Same with others.

.kiro/steering/conditional/docs/generic-report-xlsx.md
dondunne, line 1: Don't know where to put top-level comments, but here's what my Kiro complained about.

Issues / Suggestions:

MarkdownHtmlUtil.getMarkdownParserOptions() returns MutableDataSet — The Javadoc says "callers must NOT mutate" but the return type is mutable. Consider returning DataSet (immutable interface) instead, or at minimum making the field private static final and defensive-copying. As-is, any caller could inadvertently mutate the shared options.

SimpleDateFormat stored as static final in DateSelectionDialog — (From the earlier diff) SimpleDateFormat is not thread-safe. On a REST endpoint with concurrent requests, MarkdownTableRenderer creating new Parser instances is fine, but the SDF in DatePickerDialog (now DateSelectionDialog) could be problematic if ever used server-side. For the XLSX renderer's CSS and HTML rendering, this is fine since it's all fresh instances.

MarkdownTableRenderer.pdfToImage() uses deprecated PDDocument.load(InputStream) — With PDFBox 3.x, this should be Loader.loadPDF(byte[]). If you're on PDFBox 2.x it's fine, but worth noting for future migration.

trimBottomWhitespace samples every 4th pixel — This could miss thin content (e.g., a 1px border at the bottom of a table). Sampling every pixel on the last few rows might be worth it since it's only iterating one row at a time going upward and breaks on first content found.

HTML report writeTopRow vs writeHeaderRow — The top row uses <thead> open tag but doesn't close it until after the header row. This means <thead> wraps both the level names row and the column names row, which is actually correct semantically. Just confirming that's intentional.

PublishTemplateReportXlsx.writeReportData creates fonts via styles — Each createHeaderStyle(), createTopRowStyle(), etc. creates a new Font. With the WorkbookFonts already being created for markdown cells, consider sharing fonts between styles and the markdown renderer to minimize total font count.

Missing readResponseBody helper in test — The test references readResponseBody(response) but it's not shown in the diff. Presumably it already exists in the test class.

JSON escaping — The escapeJsonValue helper handles \n, \r, \t, \, " but not other control characters (e.g., \u0000–\u001F). For a report filename this is fine, but if error messages ever contain unusual characters it could produce invalid JSON. Consider using a JSON library for the async response.

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writers/PublishTemplateReportHtml.java
```diff
@@ -0,0 +1,310 @@
...
+      writer.write(".report-table td code { background-color: #F4F4F4; padding: 1px 4px; ");
+      writer.write("border-radius: 2px; font-family: Consolas, monospace; font-size: 10pt; }\n");
+      writer.write(".report-table td pre { background-color: #F4F4F4; padding: 8px; border-radius: 4px; ");
+      writer.write("font-family: Consolas, monospace; font-size: 10pt; overflow-x: auto; }\n");
+      writer.write(".report-table td img { max-width: 400px; height: auto; }\n");
+      writer.write(".report-table td blockquote { border-left: 3px solid #D9E2F3; margin: 4px 0; ");
+      writer.write("padding: 4px 12px; color: #555; }\n");
+      writer.write("</style>\n");
+   }
+
```
dondunne, line 271: Use method in AHTML

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writers/PublishTemplateReportXlsx.java
```diff
@@ -0,0 +1,237 @@
...
+   private final BranchToken branch;
+   private final ArtifactId view;
+   private final ArtifactId reportTemplateArt;
+   private final GenericReportBuilder report;
+   private final XResultData results;
+   private final MarkdownExcelCellRenderer cellRenderer;
+
+   public PublishTemplateReportXlsx(OrcsApi orcsApi, BranchId branch, ArtifactId view, ArtifactId templateArt) {
+      this.orcsApi = orcsApi;
+      this.queryApi = orcsApi.getQueryFactory();
```
dondunne: Normally not supposed to do any work in constructor.  Any reason to not do all this in a run, setup method?

plugins/org.eclipse.osee.framework.core/OSEE-INF/markdown/markdownToExcelStyles.css
david-w-miller: The markdown for flexmark is in that same directory. For example, when I searched for css:

$ find . | grep -v jsonforms | grep -v node_modules | grep -v ajax | grep -v web\/apps | grep "\.css"
./plugins/org.eclipse.osee.app/bin/org/eclipse/osee/app/html/oseeApplet.css
./plugins/org.eclipse.osee.app/OSEE-INF/oseeApp/css/app.css
./plugins/org.eclipse.osee.app/src/org/eclipse/osee/app/html/oseeApplet.css
./plugins/org.eclipse.osee.ats.rest/bin/org/eclipse/osee/ats/rest/internal/report/html/build_memo.css
./plugins/org.eclipse.osee.ats.rest/OSEE-INF/web/agileui/agile.css
./plugins/org.eclipse.osee.ats.rest/OSEE-INF/web/agileui/kanban/kanban.css
./plugins/org.eclipse.osee.ats.rest/OSEE-INF/web/agileui/kanban/ngDraggable.css
./plugins/org.eclipse.osee.ats.rest/OSEE-INF/web/agileui/program.css
./plugins/org.eclipse.osee.ats.rest/OSEE-INF/web/lib/atsCore.css
./plugins/org.eclipse.osee.ats.rest/OSEE-INF/web/lib/atsJournal.css
./plugins/org.eclipse.osee.ats.rest/OSEE-INF/web/lib/world.css
./plugins/org.eclipse.osee.ats.rest/OSEE-INF/web/ngGridExample/example.css
./plugins/org.eclipse.osee.ats.rest/src/org/eclipse/osee/ats/rest/internal/report/html/build_memo.css

./plugins/org.eclipse.osee.framework.core/OSEE-INF/markdown/markdownToExcelStyles.css
./plugins/org.eclipse.osee.framework.core/OSEE-INF/markdown/markdownToHtmlStyles.css
./plugins/org.eclipse.osee.framework.core/OSEE-INF/markdown/markdownToPdfStyles.css

./plugins/org.eclipse.osee.framework.help.ui/intro/css/overview.css
./plugins/org.eclipse.osee.framework.help.ui/intro/css/tutorials.css
./plugins/org.eclipse.osee.framework.help.ui/intro/css/whatsnew.css
./plugins/org.eclipse.osee.orcs.rest/OSEE-INF/web/applicui/config/plconfig.css
./plugins/org.eclipse.osee.orcs.rest/OSEE-INF/web/config/config.css
./plugins/org.eclipse.osee.orcs.rest/OSEE-INF/web/writer/writer.css
./plugins/org.eclipse.osee.template.engine/bin/org/eclipse/osee/template/engine/css/built-editor.css
./plugins/org.eclipse.osee.template.engine/src/org/eclipse/osee/template/engine/css/built-editor.css
./plugins/org.eclipse.osee.template.engine.test/bin/org/eclipse/osee/template/engine/css/myTest.css
./plugins/org.eclipse.osee.template.engine.test/src/org/eclipse/osee/template/engine/css/myTest.css
./plugins/org.eclipse.osee.web.ui/src/dispo/css/admin.css
./plugins/org.eclipse.osee.web.ui/src/dispo/css/user.css
./plugins/org.eclipse.osee.web.ui/src/dispo/css/userGrid.css
./plugins/org.eclipse.osee.web.ui/src/osee/css/landing.css

In the middle of that, the file added in this commit is right with the others like it. 

The Kiro response was: 
Should you consolidate? I'd say no — keep them separate, but here's the reasoning:

These stylesheets serve fundamentally different rendering targets with conflicting constraints:

HTML styles — full-featured, responsive, includes dark mode media queries, heading sizing, link colors, list styling. Designed for browser viewing.

PDF styles — constrained to 6.5in page width, fixed table layout, page-break-inside: avoid. Designed for print/PDF layout engines (OpenHTMLToPDF).

Excel styles (new) — minimal page sizing (@page { size: 8in 11in; margin: 0 }), table/code styling only. Designed to render a small fragment into a compact PNG via PDF intermediate. It intentionally omits headings, lists, links, dark mode, etc. because those don't appear in table-cell snippets.

If you merged them into one file, you'd either:

Bloat the Excel renderer with irrelevant rules (dark mode, heading sizes, responsive widths) that could interfere with the tight PNG rendering
Need conditional sections or variables that CSS doesn't naturally support in this pipeline
What I would do: Keep the three files separate (they're already in the same directory, which is good), but extract the shared base declarations into a common partial that gets composed. For example, the table, code, pre, and blockquote rules are nearly identical across all three. You could have:

OSEE-INF/markdown/
├── _base.css              ← shared table/code/blockquote rules
├── markdownToHtmlStyles.css   ← imports/includes base + adds HTML-specific (headings, dark mode, links)
├── markdownToPdfStyles.css    ← imports/includes base + adds PDF-specific (page sizing, word-wrap)
└── markdownToExcelStyles.css  ← imports/includes base + adds Excel-specific (@page, compact margins)
However, since OseeInf.getResourceContents loads a single file as a string, you'd need to concatenate at load time (no @import support in this pipeline). That's a small code change in each consumer:

String baseCss = OseeInf.getResourceContents("markdown/_base.css", ...);
String specificCss = OseeInf.getResourceContents("markdown/markdownToExcelStyles.css", ...);
String fullCss = baseCss + "\n" + specificCss;
Whether that refactor is worth it now depends on how often these styles drift apart. If you expect them to diverge further (different table styling per target), keep them independent. If you want a single source of truth for "what does an OSEE markdown table look like," the shared base approach is cleaner.

My recommendation: leave them as-is for now and note in a comment at the top of each file which target it serves. The duplication is minimal (~6 shared rules), and the output targets have legitimately different requirements. Consolidation adds complexity for marginal DRY benefit.

For the moment, I want to leave them as is/

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writers/PublishTemplateReportHtml.java
david-w-miller: I asked Kiro to clean it up - response was:
1. CSS extracted to a resource file

Created 
reportStyles.css
 — the full stylesheet lives here now as a proper .css file with normal formatting (not crammed into Java string literals)
writeStyles() now loads it via OseeInf.getResourceContents(...) — same pattern used by MarkdownConverter and MarkdownTableToPdfImage in framework.core
2. Eliminated the duplicated extensionToMimeType switch

Replaced with a one-liner that uses the existing MarkdownHtmlUtil.EXTENSION_TO_MEDIA_TYPE map — that map already defines the same mappings
3. writeErrorPage left as-is

It's 7 lines of trivial HTML wrapping a dynamic loop. Extracting it to a template file would require a placeholder/token-replacement system (or a templating engine) that doesn't exist in this bundle. The overhead isn't justified for something this small. Kiro pointed out that unlike the CSS (which is 80+ lines of static content), the error page is almost entirely dynamic (title + variable-length error list).
The streaming model is fully preserved — nothing allocates the full document in memory.

.kiro/steering/conditional/docs/generic-report-xlsx.md
david-w-miller: It gave me about the same feedback. I fixed a bunch, and preferred the current code to the suggested fixes for the above items. E.g. the MutableDataSet would require out of scope changes for this change, so I compromised with it to make a comment.

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writers/PublishTemplateReportHtml.java
david-w-miller: AHTML breaks the streaming model of this class

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/writers/PublishTemplateReportXlsx.java
david-w-miller: This constructor is just setting final values - e.g. no real work is being done, except possibly for the branch query. I have refactored the branch query out, so now it is just doing the correct constructor work.

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/publishing/markdown/MarkdownTableToPdfImage.java
david-w-miller: I set a new comment policy in Kiro. It fixed up the comments.

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/publishing/markdown/MarkdownHtmlUtil.java
david-w-miller: I set a new comment policy in Kiro. It fixed up the comments.

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/publishing/markdown/MarkdownHtmlUtil.java
david-w-miller: done

plugins/org.eclipse.osee.framework.core/src/org/eclipse/osee/framework/core/publishing/markdown/MarkdownTableToPdfImage.java
david-w-miller: done

dondunne (dismissed):

jadenpuckett (dismissed):

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/ReportEndpointImpl.java
```diff
@@ -178,6 +231,18 @@ private static int intPow(int base, int exponent) {
...
+    * Escapes a string for safe inclusion as a JSON string value. Handles double quotes, backslashes, and control
+    * characters.
+    */
+   private static String escapeJsonValue(String value) {
+      if (value == null) {
+         return "";
+      }
+      return value.replace("\\", "\\\\").replace("\"", "\\\"").replace("\n", "\\n").replace("\r", "\\r").replace("\t",
+         "\\t");
+   }
```
Grumb1e, line 254: Your method covers common cases, but it does not escape all JSON control characters—JSON requires escaping every character from U+0000 through U+001F, not only newline, carriage return, and tab.

Use a JSON library rather than hand-rolling it

```java
import com.fasterxml.jackson.databind.ObjectMapper;

private static final ObjectMapper MAPPER = new ObjectMapper();

String jsonStringLiteral = MAPPER.writeValueAsString(value);
```

Or if you want it without quotes:

```java
String escapedContents = MAPPER.writeValueAsString(value);
escapedContents = escapedContents.substring(1, escapedContents.length() - 1);
```

david-w-miller: Good comment - done

Grumb1e (approved):

jadenpuckett (approved):

PR #646, opened 2026-08-19
bug[TW29766]: IDE Client artifact does not have latest tx from rel-on…

dondunne:
…ly change

Change-Id: I5f54374c62e5e44d5662403ca80145b78a06f64c


plugins/org.eclipse.osee.orcs/src/org/eclipse/osee/orcs/data/tx_session_debug.md
jadenpuckett, line 1: i dont think we need this file

jadenpuckett (changes requested):

plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/search/engines/QueryEngineImpl.java
```diff
@@ -554,11 +564,14 @@ private void loadArtifactsInto(QueryData queryData, QueryFactory queryFactory,
                GammaId relGamma = GammaId.valueOf(stmt.getLong("gamma_id"));
                artifact.putRelation(tokenService.getRelationTypeOrCreate(typeId), side, otherArtifact, relGamma);
             }
+            // Update artifact's transaction if this relation row has a newer transaction
+            TransactionId relTxId = TransactionId.valueOf(stmt.getLong("transaction_id"));
+            artifact.updateTransactionIfNewer(relTxId);
             if (OptionsUtil.getIncludeLatestTransactionDetails(queryData.getRootQueryData().getOptions())) {
+               artifact.updateTransactionIfNewer(suppTxId);
 
-               if (artifact.getLatestTxDetails().getAuthor().isInvalid()) {
```
audrey-denk, line 559: I wonder why this was here to begin with... removing it may be fine, just worried that it was there for a reason... I should have had better notes here I'm sure.

audrey-denk (approved):
Looks ok.  I had one comment on the removal of a line that I'm not sure about.  Being that this changes QueryEngineImpl it is a very potentially breaking change.  I would check with those folks doing the nightly JIRA syncs which rely heavily on the latest tx values to make sure everything still works as expected.

jadenpuckett (approved):

PR #648, opened 2026-08-24
feature[TW29566]: search tag updateSearch tag update

audrey-denk:
Update search tagging to use new hashing scheme


Grumb1e (dismissed):

jadenpuckett (dismissed):

Grumb1e (dismissed):

docs/hybrid-fulltext-search-migration.md
```diff
@@ -0,0 +1,174 @@
...
+The bit-packing encoder was replaced with FNV-1a 64-bit hashing:
+- One word = one hash = one tag
+- No collisions in practice (64-bit hash space)
+- Case-insensitive (normalizes to lowercase before hashing)
+
+## Gradual Rollout: Dual-Table Transition
+
+Because the release, release-candidate, and nightly tracks share a single Oracle database,
+the migration cannot be done as a single cutover. The approach:
+
```
dondunne, line 28: can you ask ai/kiro if there's anything else we should do to improve new hash table or architecture while we're at it.  Like should the table be some sort of new fangled table or new indexes or anything...

docs/hybrid-fulltext-search-migration.md
```diff
@@ -0,0 +1,174 @@
...
+- `osee_search_tags` — kept intact. The legacy `release` app server continues writing
+  bit-packed tags here and querying from here. No changes to release code.
+- `osee_search_tags_hash` — new table with identical schema. The `nightly`/`rc` app servers
+  write hash-based tags here and query from here.
+
```
dondunne, line 33: I don't want to mass change the types cause of the Widget changes, however, we after I merge we should remove the tag option from types and only have a no-tag option if we know it shouldn't be tagged.  Is it easy to run a update after that to tag all missing?

If you agree, can we write an action to update types and update tags after Widget merge.  Or I can write if you confirm that's a good thing and it's easy to do after.

docs/hybrid-fulltext-search-migration.md
```diff
@@ -0,0 +1,174 @@
...
+The nightly Java code dual-writes during normal attribute indexing: every tag written by the
+`IndexingTaskDatabaseTxCallable` goes to both tables, using the appropriate encoding for each
+(legacy bit-packed tags for `osee_search_tags`, FNV-1a hash tags for `osee_search_tags_hash`).
+This covers attributes indexed while the nightly server is running. However, attributes created
+by the `release` server are only indexed into `osee_search_tags`. A nightly CI job bridges this gap.
```
dondunne, line 38: What happens to our local dev environments, demo db, runtime and etc?  Will they only use new tagging table after dbinit?  Do we have to coordinate with developers to re-dbinit or will it handle the new table not being there??

docs/hybrid-fulltext-search-migration.md
```diff
@@ -0,0 +1,174 @@
...
+For the initial migration (empty `osee_search_tags_hash` table), use the async bulk endpoint:
+
+```
+POST /orcs/index/reindex/direct
+```
+
+Returns 202 immediately. Indexes all current taggable attributes missing from the hash table.
+Server-side concurrency is bounded to 2 threads via a shared thread pool, preventing database
+overload regardless of how many requests are submitted.
+
```
dondunne, line 74: Create action for this Cutover checklist

plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/search/indexer/QueryEngineIndexerImpl.java
```diff
@@ -110,31 +131,239 @@ public void indexAttrTypeIds(OrcsSession session, OrcsTokenService tokenService,
...
+         // Wait for all indexing tasks to complete before returning
+         for (Future<?> future : indexFutures) {
+            future.get(10, TimeUnit.MINUTES);
+         }
       } catch (Exception ex) {
          OseeCoreException.wrapAndThrow(ex);
       }
+      logger.info("Completed tagging for attr type %d", attrTypeId);
+   }
+
```
dondunne, line 199: Think we talked about adding an endpoint for indexing by artid or list of artids?

plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/search/tagger/LegacyTagEncoder.java
```diff
@@ -0,0 +1,65 @@
...
+package org.eclipse.osee.orcs.db.internal.search.tagger;
+
+import java.util.Arrays;
+
+/**
+ * Original bit-packing encoder preserved for dual-write compatibility during the transition period. Produces the same
+ * multi-row tags that the legacy release search handler expects to find in osee_search_tags.
+ * <p>
+ * Remove this class once all release tracks have migrated to hash-based search via osee_search_tags_hash.
+ *
```
dondunne, line 23: Author? (and rest of new files)

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/IndexerEndpointImpl.java
```diff
@@ -108,4 +125,90 @@ public Response deleteIndexQueue() {
...
+      if (attrTypeId > 0) {
+         attrTypeIds.add(attrTypeId);
+      } else {
+         for (var attrType : orcsApi.tokenService().getTaggedAttrs()) {
+            attrTypeIds.add(attrType.getId());
+         }
+      }
+
+      int typeCount = attrTypeIds.size();
+
```
dondunne, line 197: use /** * */ and let formatter wrap

dondunne (commented):
Kiro Review

I'd do 1, 3 and 5.  Don't understand 2 or 4, but maybe you do.

---

PR #648: Search Tag Update — Review
Author: audrey-denk | Status: Open, 1 approval (Grumb1e) | Branch: searchTagUpdate → main

Summary
Replaces the broken bit-packing TagEncoder (which had collisions due to 5-bit values crammed into 4-bit slots) with FNV-1a 64-bit hashing. Introduces a dual-table transition strategy so the legacy release track keeps working on osee_search_tags while nightly/RC uses the new osee_search_tags_hash table. Adds REST endpoints for bulk and incremental re-indexing plus thorough migration documentation.

The design is solid and well-documented. The dual-write approach is the right call given the shared database constraint. Five issues worth addressing before merge:

Issues
1. InputStream leak on tagger exception (confirmed)

In both indexDirectByAttrType and indexRecentlyModified:

java

InputStream input = getInputStream(value, uri);
if (input != null) {
    tagger.tagIt(input, collector);
    input.close();  // skipped if tagIt() throws
}
If tagger.tagIt() throws, the stream is never closed. For URI-backed resources, this leaks OS file handles. Over a bulk re-index of thousands of attributes, this can exhaust file descriptors. Use try-with-resources.

2. Silent failure in async reindex tasks (confirmed)

REINDEX_EXECUTOR.submit() discards the returned Future. If the background task throws (JDBC timeout, constraint violation, OOM), nothing is logged. Since the 202 response already told the caller "it's running," server logs are the only feedback mechanism. Wrap in try { ... } catch (Exception ex) { logger.error(...); }.

3. Unbounded hours parameter on /reindex/recent (confirmed)

No server-side cap on the hours value. A large number (e.g., 999999) effectively removes the time filter, and the hard-coded Oracle hint (/*+ LEADING(txd txs att oji) USE_NL(txs att) */) assumes a narrow time window. Combined, this can force nested-loop joins on a huge rowset. Add a ceiling like Math.min(hours, 168) or reject values above a threshold.

4. Partial-index risk on crash recovery (edge case)

The NOT EXISTS guard checks whether any row exists for a gamma, not whether all tags are present. If a batch insert partially succeeds (some tags written before a crash), the gamma is considered "indexed" on retry even though its tag set is incomplete. INTERSECT queries would then fail to find the document. Low risk since JDBC batch ops tend to be atomic, but INSERT ... ON CONFLICT DO NOTHING or a per-gamma marker would make this bulletproof.

5. Flush-threshold asymmetry in dual-write (minor)

isStorageAllowed(hashTagsToStore) only checks the hash map. The legacy map grows faster (multiple tags per word due to 12-char chunking) and accumulates unbounded between hash-triggered flushes. Unlikely to cause memory pressure in normal workloads, but worth noting for large batch scenarios.

SpotBugs Findings
The CI SpotBugs report flagged ~15 issues. Most are pre-existing EI_EXPOSE_REP2 patterns (storing mutable objects) not introduced by this PR. The two findings specific to this PR — ignored Future return values in reindexAllCurrent and reindexDirect — align with issue #2 above.

Verdict
Approve with minor fixes. Issues 1–3 are worth fixing before merge (resource leak, silent failures, unbounded query). Issues 4–5 are acceptable to track as follow-ups. The overall architecture, migration strategy, and documentation quality are strong.

david-w-miller:
I ran it through an independent kiro review - it had some suggestions. I will forward those to you via email.

jadenpuckett (dismissed):

docs/hybrid-fulltext-search-migration.md
audrey-denk: Kiro actually came up with the table.  It does have a couple indexes and seems to perform well.

docs/hybrid-fulltext-search-migration.md
audrey-denk: yes it should be easy enough to do after the widget merge.

docs/hybrid-fulltext-search-migration.md
audrey-denk: yes they will have to re-dbinit... I assumed most people did that, but good point.  That will need to be communicated.

docs/hybrid-fulltext-search-migration.md
audrey-denk: Do you mean for final?  Like when this makes its way to release?  Will do.

plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/search/indexer/QueryEngineIndexerImpl.java
audrey-denk: We have one for indexing by a specific type, but not a specific artid.  That would be easy enough to add though.

plugins/org.eclipse.osee.orcs.db/src/org/eclipse/osee/orcs/db/internal/search/tagger/LegacyTagEncoder.java
audrey-denk: Can't we just get that from history?

Grumb1e (approved):

jadenpuckett (approved):

PR #650, opened 2026-08-27
feature[TW29769]: PR - Create new Reqts Check Signals feature

dondunne:


david-w-miller (commented):
My Kiro review might be different than yours, here is what it said:
Code Review: Signal Checker Feature (ICD)
Summary
This is a well-structured initial implementation of the signal checker feature following the architecture described in the steering document. The code is cleanly layered (API → Core → Server/IDE), properly delegates across the client-server boundary, and the test infrastructure is solid. Below are findings grouped by severity.

Errors (Potential bugs / correctness issues)
1. SignalCheckerData — Jackson serialization mismatch on artSigData
File: 
SignalCheckerData.java
 (lines 37, 136-142)

The getter is getArtSigDatas() (plural 's') but the setter is setArtSigData() (no trailing 's') and the field is artSigData. Jackson derives the JSON property name from the getter/setter pair. When they disagree, Jackson will:

Serialize using getArtSigDatas() → JSON key "artSigDatas"
Deserialize expecting setArtSigDatas() or fall back to setArtSigData() → JSON key "artSigData"
This means the round-trip serialize → deserialize will fail: the server returns "artSigDatas": [...] in JSON, but the client-side Jackson instance cannot map it back because the setter suggests "artSigData". The test might pass now because it runs in a single JVM, but a real REST round-trip over HTTP will lose the artifact signal data list.

Fix: Rename either the getter to getArtSigData() or the setter to setArtSigDatas() (and the field accordingly) so they agree.

2. CheckSignalsIcdTest — error count assertion is fragile
File: 
CheckSignalsIcdTest.java
 (line 35)

Assert.assertEquals(2, result.getRd().getErrorCount());
The checkSignalsTest implementation logs errors via rd.errorf(...) which bumps errorCount, but also calls scd.setTotalErrors(scd.getTotalErrors() + 1). Counting through the code:

Phase 5: BAD SIGNAL!@# → 1 rd.errorf()
Phase 6: EMITTERS.E(01..05)_MODE → 1 rd.errorf()
That gives 2, which is correct for the current implementation. However, if you look closely at Phase 6 logic, the test assertion depends on the simulated missing list (EMITTERS.E03_MODE, EMITTERS.E05_MODE) being treated as a single error per SD (not per missing signal). This is correct as designed, but the assertion assertEquals(2, ...) doesn't document why it expects 2. Consider adding a comment or asserting more semantic properties (e.g., totalUndefinedSignals == 2).

Severity: Warning (correctness is fine now, but the test is brittle to implementation changes)

Warnings (Potential issues / code quality)
3. SignalCheckerData.artsTocheck — naming inconsistency
File: 
SignalCheckerData.java
 (line 31)

The field is artsTocheck with a lowercase 'c' in "check". Java convention for camelCase would be artsToCheck. Since this is a serializable POJO used in REST APIs, the JSON property name will be "artsTocheck". This is a minor inconsistency but worth noting since once shipped, renaming the JSON field becomes a breaking change.

4. SigDbDemoImpl.populateDemoSignals() — not thread-safe lazy initialization
File: 
SigDbDemoImpl.java
 (lines 36-56)

private void populateDemoSignals() {
   if (signals.isEmpty()) {
      // ...populate...
   }
}
The check-then-act on signals.isEmpty() is not synchronized. If two threads call isSignalDefined concurrently before the map is populated, you could get partial initialization or duplicate entries. For a demo/test class this is low risk, but it's a pattern to avoid.

Suggestion: Either populate in the constructor, or use a volatile boolean guard or synchronize the method.

5. SigDbDemoImpl.readSignal() — does not call populateDemoSignals()
File: 
SigDbDemoImpl.java
 (line 68)

@Override
public SignalDefinition readSignal(String signalName, String variantId) {
   return signals.get(signalName);
}
Unlike isSignalDefined(), readSignal() does not call populateDemoSignals() first. If readSignal is called before isSignalDefined, it will always return null.

Fix: Add populateDemoSignals() at the start of readSignal().

6. SigDbDemoImpl — variantId parameter is ignored
File: 
SigDbDemoImpl.java
 (lines 63-68)

Both isSignalDefined and readSignal ignore the variantId parameter entirely. The class advertises "DEMO_V1" as its only variant, but doesn't validate that the caller passes this variant. Passing a non-existent variant will still return results, which violates the contract implied by listVariantIds().

Suggestion: Add a guard: if (!variantIds.contains(variantId)) return false / null.

7. SignalExpander.expandRange — potential overflow with large ranges
File: 
SignalExpander.java
 (line 137)

if (end - start > 9999) {
The check prevents ranges larger than 10,000 values, which is good. However, end - start can overflow if start is very large and end is Integer.MAX_VALUE. Since the inputs are zero-padded strings (typically small numbers), this is unlikely in practice, but worth noting.

8. AtsIcdServiceServerImpl.checkSignalsTest() — double-counting errors
File: 
AtsIcdServiceServerImpl.java

The method calls both rd.errorf(...) (which increments XResultData.errorCount) and scd.setTotalErrors(scd.getTotalErrors() + 1) (which increments the model's totalErrors). These are two different counters tracking the same concept but independently. They will drift apart if any phase logs an error without updating totalErrors or vice versa.

Suggestion: Consider deriving totalErrors from rd.getErrorCount() at the end of the run, or only using one tracking mechanism.

Suggestions (Style / maintainability)
9. SignalData.severity — should be an enum, not a String
File: 
SignalData.java

The severity field uses magic strings "OK", "ERROR", and empty string. The architecture document mentions "severity enums" as part of the constants. Using an enum (e.g., SignalSeverity.OK, SignalSeverity.ERROR, SignalSeverity.WARNING) would prevent typos and enable exhaustive switch statements.

10. SignalCheckerHtmlReport — XSS concern with raw signal names
File: 
SignalCheckerHtmlReport.java
 (line 68)

AHTML.bold(sd.getRawSignal())
If a rawSignal contains HTML special characters (e.g., <, >, &), they will be rendered as HTML. The test data actually includes BAD SIGNAL!@# which is safe, but a signal like <script>alert(1)</script> would be problematic if this HTML is rendered in a browser context. AHTML.textToHtml() should be used for user-provided content, or the signal name should be HTML-escaped.

11. CheckSignalsRoundTripTestNavigateItem — no error handling on REST failure
File: 
CheckSignalsRoundTripTestNavigateItem.java

The run() method calls the REST endpoint without any try/catch. If the server is unavailable or the call fails, the user gets an unhandled exception dialog. Other navigate items in the codebase typically wrap REST calls with error handling.

12. Unused import in AtsApiIdeImpl
File: 
AtsApiIdeImpl.java

The import org.eclipse.osee.ats.ide.reqts.icd.AtsIcdServiceIdeImpl appears before org.eclipse.osee.ats.api.query.IAtsQueryServiceServer. Based on the typical import organization (API before implementation classes), the placement is fine but verify that your Eclipse formatter/import organizer handles the ordering.

13. AtsIcdEndpointApi missing @Path annotation completeness
File: 
AtsIcdEndpointApi.java

The endpoint declares @Path("icd") at the interface level. Since AtsApplication is registered at @ApplicationPath("ats"), the full paths will be /ats/icd/chksig and /ats/icd/chksigtest — which matches the architecture doc. This is correct. No issue.

Overall Assessment
The implementation is solid and well-aligned with the architecture steering document. The most impactful issue is #1 (the getter/setter name mismatch) which will cause JSON deserialization failures on actual REST round-trips. Issue #5 (missing populateDemoSignals() call) is a straightforward bug in the demo implementation. The rest are code quality improvements worth addressing before this ships to production.

dondunne:
> My Kiro review might be different than yours, here is what it said: Code Review: Signal Checker Feature (ICD) Summary This is a well-structured initial implementation of the signal checker feature following the architecture described in the steering document. The code is cleanly layered (API → Core → Server/IDE), properly delegates across the client-server boundary, and the test infrastructure is solid. Below are findings grouped by severity.
> 
> Errors (Potential bugs / correctness issues)
> 
> 1. SignalCheckerData — Jackson serialization mismatch on artSigData
>    File:
>    SignalCheckerData.java
>    (lines 37, 136-142)
> 
> The getter is getArtSigDatas() (plural 's') but the setter is setArtSigData() (no trailing 's') and the field is artSigData. Jackson derives the JSON property name from the getter/setter pair. When they disagree, Jackson will:
> 
> Serialize using getArtSigDatas() → JSON key "artSigDatas" Deserialize expecting setArtSigDatas() or fall back to setArtSigData() → JSON key "artSigData" This means the round-trip serialize → deserialize will fail: the server returns "artSigDatas": [...] in JSON, but the client-side Jackson instance cannot map it back because the setter suggests "artSigData". The test might pass now because it runs in a single JVM, but a real REST round-trip over HTTP will lose the artifact signal data list.
> 
> Fix: Rename either the getter to getArtSigData() or the setter to setArtSigDatas() (and the field accordingly) so they agree.
> 
> 2. CheckSignalsIcdTest — error count assertion is fragile
>    File:
>    CheckSignalsIcdTest.java
>    (line 35)
> 
> Assert.assertEquals(2, result.getRd().getErrorCount()); The checkSignalsTest implementation logs errors via rd.errorf(...) which bumps errorCount, but also calls scd.setTotalErrors(scd.getTotalErrors() + 1). Counting through the code:
> 
> Phase 5: BAD SIGNAL!@# → 1 rd.errorf() Phase 6: EMITTERS.E(01..05)_MODE → 1 rd.errorf() That gives 2, which is correct for the current implementation. However, if you look closely at Phase 6 logic, the test assertion depends on the simulated missing list (EMITTERS.E03_MODE, EMITTERS.E05_MODE) being treated as a single error per SD (not per missing signal). This is correct as designed, but the assertion assertEquals(2, ...) doesn't document why it expects 2. Consider adding a comment or asserting more semantic properties (e.g., totalUndefinedSignals == 2).
> 
> Severity: Warning (correctness is fine now, but the test is brittle to implementation changes)
> 
> Warnings (Potential issues / code quality) 3. SignalCheckerData.artsTocheck — naming inconsistency File: SignalCheckerData.java (line 31)
> 
> The field is artsTocheck with a lowercase 'c' in "check". Java convention for camelCase would be artsToCheck. Since this is a serializable POJO used in REST APIs, the JSON property name will be "artsTocheck". This is a minor inconsistency but worth noting since once shipped, renaming the JSON field becomes a breaking change.
> 
> 4. SigDbDemoImpl.populateDemoSignals() — not thread-safe lazy initialization
>    File:
>    SigDbDemoImpl.java
>    (lines 36-56)
> 
> private void populateDemoSignals() { if (signals.isEmpty()) { // ...populate... } } The check-then-act on signals.isEmpty() is not synchronized. If two threads call isSignalDefined concurrently before the map is populated, you could get partial initialization or duplicate entries. For a demo/test class this is low risk, but it's a pattern to avoid.
> 
> Suggestion: Either populate in the constructor, or use a volatile boolean guard or synchronize the method.
> 
> 5. SigDbDemoImpl.readSignal() — does not call populateDemoSignals()
>    File:
>    SigDbDemoImpl.java
>    (line 68)
> 
> @OverRide public SignalDefinition readSignal(String signalName, String variantId) { return signals.get(signalName); } Unlike isSignalDefined(), readSignal() does not call populateDemoSignals() first. If readSignal is called before isSignalDefined, it will always return null.
> 
> Fix: Add populateDemoSignals() at the start of readSignal().
> 
> 6. SigDbDemoImpl — variantId parameter is ignored
>    File:
>    SigDbDemoImpl.java
>    (lines 63-68)
> 
> Both isSignalDefined and readSignal ignore the variantId parameter entirely. The class advertises "DEMO_V1" as its only variant, but doesn't validate that the caller passes this variant. Passing a non-existent variant will still return results, which violates the contract implied by listVariantIds().
> 
> Suggestion: Add a guard: if (!variantIds.contains(variantId)) return false / null.
> 
> 7. SignalExpander.expandRange — potential overflow with large ranges
>    File:
>    SignalExpander.java
>    (line 137)
> 
> if (end - start > 9999) { The check prevents ranges larger than 10,000 values, which is good. However, end - start can overflow if start is very large and end is Integer.MAX_VALUE. Since the inputs are zero-padded strings (typically small numbers), this is unlikely in practice, but worth noting.
> 
> 8. AtsIcdServiceServerImpl.checkSignalsTest() — double-counting errors
>    File:
>    AtsIcdServiceServerImpl.java
> 
> The method calls both rd.errorf(...) (which increments XResultData.errorCount) and scd.setTotalErrors(scd.getTotalErrors() + 1) (which increments the model's totalErrors). These are two different counters tracking the same concept but independently. They will drift apart if any phase logs an error without updating totalErrors or vice versa.
> 
> Suggestion: Consider deriving totalErrors from rd.getErrorCount() at the end of the run, or only using one tracking mechanism.
> 
> Suggestions (Style / maintainability) 9. SignalData.severity — should be an enum, not a String File: SignalData.java
> 
> The severity field uses magic strings "OK", "ERROR", and empty string. The architecture document mentions "severity enums" as part of the constants. Using an enum (e.g., SignalSeverity.OK, SignalSeverity.ERROR, SignalSeverity.WARNING) would prevent typos and enable exhaustive switch statements.
> 
> 10. SignalCheckerHtmlReport — XSS concern with raw signal names
>     File:
>     SignalCheckerHtmlReport.java
>     (line 68)
> 
> AHTML.bold(sd.getRawSignal()) If a rawSignal contains HTML special characters (e.g., <, >, &), they will be rendered as HTML. The test data actually includes BAD SIGNAL!@# which is safe, but a signal like <script>alert(1)</script> would be problematic if this HTML is rendered in a browser context. AHTML.textToHtml() should be used for user-provided content, or the signal name should be HTML-escaped.
> 
> 11. CheckSignalsRoundTripTestNavigateItem — no error handling on REST failure
>     File:
>     CheckSignalsRoundTripTestNavigateItem.java
> 
> The run() method calls the REST endpoint without any try/catch. If the server is unavailable or the call fails, the user gets an unhandled exception dialog. Other navigate items in the codebase typically wrap REST calls with error handling.
> 
> 12. Unused import in AtsApiIdeImpl
>     File:
>     AtsApiIdeImpl.java
> 
> The import org.eclipse.osee.ats.ide.reqts.icd.AtsIcdServiceIdeImpl appears before org.eclipse.osee.ats.api.query.IAtsQueryServiceServer. Based on the typical import organization (API before implementation classes), the placement is fine but verify that your Eclipse formatter/import organizer handles the ordering.
> 
> 13. AtsIcdEndpointApi missing @path annotation completeness
>     File:
>     AtsIcdEndpointApi.java
> 
> The endpoint declares @path("icd") at the interface level. Since AtsApplication is registered at @ApplicationPath("ats"), the full paths will be /ats/icd/chksig and /ats/icd/chksigtest — which matches the architecture doc. This is correct. No issue.
> 
> Overall Assessment The implementation is solid and well-aligned with the architecture steering document. The most impactful issue is #1 (the getter/setter name mismatch) which will cause JSON deserialization failures on actual REST round-trips. Issue #5 (missing populateDemoSignals() call) is a straightforward bug in the demo implementation. The rest are code quality improvements worth addressing before this ships to production.

Fixed 1 and 3, rest are minor or will be improved in later commits as this feature matures.

.kiro/steering/conditional/docs/reqts_signal_checker_architecture.md
```diff
@@ -0,0 +1,441 @@
+# Mission Software Requirements Signal Checking
+
+## Signal Checking Description
+
```
Grumb1e: There is no inclusion mechanism, and the main content of the docs should go in docs/ai

The kiro inclusion mechanism should go in this file, but the actual content should got in docs/ai.

That way the information is discoverable by Codex or other agents as well. 

There are a few other cases like this that may be worth fixing now:
.kiro\steering\conditional\docs\ide_build-impact-table-architecture.md
.kiro\steering\conditional\docs\ide_event-system-architecture.md
.kiro\steering\conditional\docs\ide_xviewer-architecture.md

dondunne: Done

Grumb1e (approved):

david-w-miller (approved):

PR #651, opened 2026-08-27
feature[TW29758]: Create initial export of ATS CM to SysML

dondunne:
Change-Id: Iff9a1ba4ce3c48c960a99bfffae140345735af36


david-w-miller:
Kiro review wins again: 
Code Review: ATS Program to SysML V2 Export
Summary
This change introduces a new feature to export ATS Program configuration and workflow objects to SysML V2 textual notation. It adds a Java model layer (org.eclipse.osee.ats.api.sysml package), two exporters (ConfigExporter, WorkflowExporter), a text serializer (SysmlTextWriter), REST endpoints, and an IDE navigate item.

DemoArtifactToken.java
[suggestion] The specific imports were replaced with a wildcard import static org.eclipse.osee.ats.api.data.AtsArtifactTypes.*. Per the project's import standards, wildcard imports are appropriate when 5+ imports come from the same package (there were 5 here). This is acceptable.

OseeTypeToSysmlMapper.java
[warning] Line ~59: toEnumDef calls attrType.toEnum() without null safety

If toEnum() returns null on an unexpected type, the next call to enumType.getEnumStrValues() will NPE. Although the preceding isEnumerated() check should prevent this, the method is public and could be called from other contexts.

[suggestion] Line ~67-82: toSysmlIdentifier drops the capitalizeNext flag when the first character is a separator

If the input starts with separators (e.g., " foo"), the result starts with an empty string and the first real character won't be lowercased properly. The i == 0 check on the else branch handles only the first input character, not the first output character. Edge case, but worth noting.

[warning] Line ~74: Missing handling for consecutive separators

Input like "foo--bar" silently swallows the double-dash with capitalizeNext remaining true, which is correct. But input consisting entirely of separator characters (e.g., "---") produces an empty string, falling through to the id.isEmpty() → "_" guard. This works, but the contract isn't documented.

OseeRelationToSysmlMapper.java
[error] Line ~57-72: switch uses default fallthrough but missing UNORDERED case

RelationTypeMultiplicity likely includes UNORDERED or other values beyond the four listed. The default case handles this, but the switch should use an exhaustive pattern or an explicit comment noting why other values get [*].

[suggestion] Lines ~81-128: Three very similar string conversion methods (toSysmlRelationName, toEndName, toTypeName)

These could be consolidated into a single parameterized method with a flag for PascalCase vs camelCase. Not critical but reduces maintenance burden.

SysmlTextWriter.java
[error] Line ~161: Unchecked cast (List<Object>) attrVal.getValue()

The @SuppressWarnings("unchecked") annotation suppresses this warning, but the cast is genuinely unsafe. If getValue() returns a List<String> or List<Integer>, the cast succeeds at runtime (type erasure), but the code assumption that all list elements can be passed to formatValue(Object, ...) is fragile. Consider typing SysmlAttributeValue.getValue() more strictly or adding a runtime type check.

[warning] Line ~204: sysmlName with null/empty returns '' (two single quotes)

This could produce invalid SysML if a name is unexpectedly null. A consumer calling sysmlName(null) gets '' which may not be a valid SysML identifier in all positions.

[suggestion] Line ~183: formatValue treats null as the literal string "null"

If null values should be excluded rather than emitted, this should be filtered upstream in writeAttributeValue. Currently, a null value produces attribute x = null; which is not valid SysML V2. The ConfigExporter.addIfValid() method already filters nulls for some attributes, but not all paths guard against this.

SysmlConstants.java
[suggestion] MULTIPLICITY_ONE is an empty string

The comment says "implied [1]" which matches the SysML spec, but using an empty string as a sentinel is somewhat fragile. hasMultiplicity() in SysmlAttributeDef checks != null && !isEmpty(), so this works, but a constant named MULTIPLICITY_ONE that equals "" is counter-intuitive to future maintainers. A brief comment on the constant itself would help.

ConfigExporter.java
[warning] Line ~67-71: teamDef.getActionableItems() is called twice

Once during initial collection (line ~67) and again in the connections loop (line ~110). If getActionableItems() makes a server call or produces different results between invocations, this is a bug. If it's cached, it's just redundant work. Consider storing the result during the first pass.

[warning] Line ~73-76: getLeads and getMembers called twice per team def

Same issue as above — these are called during collection (lines 68-69) and again in the connections loops (lines 117-127). This doubles the API calls for every team definition.

[suggestion] Line ~85: addConfigEnumDefs only adds BuildImpactState

Per the design doc, enum defs should be generated from enumerated attribute types (e.g., ChangeType, WorkType, ClosureState). Currently only BuildImpactState is hard-coded. This appears to be an incomplete implementation for Phase 1 — consider adding a TODO comment.

[suggestion] Lines ~145-184: Part def construction is entirely hard-coded

The design doc states that attributes should be "derived from AtsArtifactTypes definitions" via type introspection. The current implementation manually lists attributes. This is pragmatic for Phase 1 but diverges from the stated approach. A comment acknowledging this would clarify intent.

WorkflowExporter.java
[error] Line ~75-83: The query collects all team workflows but doesn't filter by state

For large programs with thousands of completed workflows, this query could return an enormous dataset without pagination or limits. The REST endpoint has no query parameters to control scope (e.g., only open workflows, date range). This is a potential performance and memory issue.

[warning] Line ~130-137: BID connections only check isProblemReport() workflows

The BuildImpactDataToTeamWf connection (BIT → CR direction) is never emitted. The code only outputs ProblemReportToBid (PR → BIT), but per the design doc, the BIT→CR connections (BuildImpactDataToTeamWf) should also be exported by iterating BID artifacts and their related CRs.

[warning] Line ~113-114: Action-to-workflow connection uses action.getId() as source name

The SysmlConnection source name is String.valueOf(action.getId()), but actions' part usages are also named by their ID (line ~176). This is consistent but uses numeric IDs as SysML part names, which are valid but not human-readable. The design doc examples use 'ACTION_12345' format with the ATS ID.

[suggestion] Line ~154: State machine name uses wd.getName() which may not be unique

If two team defs share the same WorkDefinition object, the Set<WorkDefinition> correctly deduplicates. But if different WorkDefinition objects have the same name, both get added with duplicate state machine names in the output.

[suggestion] Lines ~97-106: Connection defs are added unconditionally

All connection defs (Derive, ResolvedBy, etc.) are emitted in the output even if no connections of that type exist in the data. While not incorrect, it adds noise. Consider only adding connection defs that have corresponding connection instances.

ExportProgramToSysmlNavigateItem.java
[warning] Line ~35-36: Hard-coded output directory

System.getProperty("user.home") + "/SawPlProgramSysMl" is not configurable. There's a TODO comment acknowledging this. Per Java coding standards: "No hard-coded absolute file paths in production code."

[warning] Line ~49-50: outputDir.mkdirs() return value ignored

mkdirs() returns false if the directory couldn't be created. The code should check this and report failure.

[error] Line ~54: Hard-coded to DemoArtifactToken.SAW_PL_Program

This navigate item is registered in the general ATS navigation view (not a demo-only view), but it only works with the SAW PL demo program. Users without demo data will get a confusing error. Either gate registration behind a demo check, or provide a program selector dialog.

[suggestion] Line ~57: No error handling for Lib.writeStringToFile

If the write fails, the exception is caught by the outer catch block, but the error message will be generic. Consider catching IOException specifically for file operations.

AtsActionEndpointApi.java (new endpoints)
[suggestion] Lines ~263-270: Missing @Consumes annotation on GET endpoints

The getSysmlConfig and getSysmlWorkflows endpoints are GETs producing TEXT_PLAIN. This is fine for the response, but unusually, these take an ArtifactId path param. No issues here, but consider whether JSON output with APPLICATION_JSON would be more useful for programmatic consumers (text is fine for file download).

AtsActionEndpointImpl.java (new endpoints)
[suggestion] Lines ~917-930: No error handling for export failures

If ConfigExporter.export() or WorkflowExporter.export() throws (e.g., program not found), the exception will propagate as an unhandled 500 error. Consider catching OseeArgumentException and returning a proper 404/400 response.

General Observations
Positive aspects:

Clean separation of concerns: model classes, mappers, exporters, serializer
Follows the project's file header and @author tag conventions correctly
Immutable model objects (final fields, no setters on most classes)
The SysmlTextWriter is well-structured and produces readable output
Good alignment with the design doc's flat-objects + connections approach
Across the change:

No unit tests are included in this diff. Given the toSysmlIdentifier, toSysmlRelationName, toEndName, toTypeName, and SysmlTextWriter logic, these are prime candidates for unit testing.
The SysmlEnumDef in ConfigExporter.addConfigEnumDefs() is hard-coded rather than generated from the type system — this should eventually use OseeTypeToSysmlMapper.toEnumDef().
Unused import: SysmlConnectionDef is imported in ConfigExporter (line ~22) and used only indirectly via the mapper return value, which is correct — disregard, the import is used.
StateType is imported but unused in WorkflowExporter.java — verify if this is dead.

dondunne:
> DemoArtifactToken.java [suggestion] The specific imports were replaced with a wildcard import static org.eclipse.osee.ats.api.data.AtsArtifactTypes.*. Per the project's import standards, wildcard imports are appropriate when 5+ imports come from the same package (there were 5 here). This is acceptable.
> 
> OseeTypeToSysmlMapper.java [warning] Line ~59: toEnumDef calls attrType.toEnum() without null safety
> 
> If toEnum() returns null on an unexpected type, the next call to enumType.getEnumStrValues() will NPE. Although the preceding isEnumerated() check should prevent this, the method is public and could be called from other contexts.
> 
> [suggestion] Line ~67-82: toSysmlIdentifier drops the capitalizeNext flag when the first character is a separator
> 
> If the input starts with separators (e.g., " foo"), the result starts with an empty string and the first real character won't be lowercased properly. The i == 0 check on the else branch handles only the first input character, not the first output character. Edge case, but worth noting.
> 
> [warning] Line ~74: Missing handling for consecutive separators
> 
> Input like "foo--bar" silently swallows the double-dash with capitalizeNext remaining true, which is correct. But input consisting entirely of separator characters (e.g., "---") produces an empty string, falling through to the id.isEmpty() → "_" guard. This works, but the contract isn't documented.
> 
> OseeRelationToSysmlMapper.java [error] Line ~57-72: switch uses default fallthrough but missing UNORDERED case
> 
> RelationTypeMultiplicity likely includes UNORDERED or other values beyond the four listed. The default case handles this, but the switch should use an exhaustive pattern or an explicit comment noting why other values get [*].
> 
> [suggestion] Lines ~81-128: Three very similar string conversion methods (toSysmlRelationName, toEndName, toTypeName)
> 
> These could be consolidated into a single parameterized method with a flag for PascalCase vs camelCase. Not critical but reduces maintenance burden.
> 
> SysmlTextWriter.java [error] Line ~161: Unchecked cast (List) attrVal.getValue()
> 
> The @SuppressWarnings("unchecked") annotation suppresses this warning, but the cast is genuinely unsafe. If getValue() returns a List or List, the cast succeeds at runtime (type erasure), but the code assumption that all list elements can be passed to formatValue(Object, ...) is fragile. Consider typing SysmlAttributeValue.getValue() more strictly or adding a runtime type check.
> 
> [warning] Line ~204: sysmlName with null/empty returns '' (two single quotes)
> 
> This could produce invalid SysML if a name is unexpectedly null. A consumer calling sysmlName(null) gets '' which may not be a valid SysML identifier in all positions.
> 
> [suggestion] Line ~183: formatValue treats null as the literal string "null"
> 
> If null values should be excluded rather than emitted, this should be filtered upstream in writeAttributeValue. Currently, a null value produces attribute x = null; which is not valid SysML V2. The ConfigExporter.addIfValid() method already filters nulls for some attributes, but not all paths guard against this.
> 
> SysmlConstants.java [suggestion] MULTIPLICITY_ONE is an empty string
> 
> The comment says "implied [1]" which matches the SysML spec, but using an empty string as a sentinel is somewhat fragile. hasMultiplicity() in SysmlAttributeDef checks != null && !isEmpty(), so this works, but a constant named MULTIPLICITY_ONE that equals "" is counter-intuitive to future maintainers. A brief comment on the constant itself would help.
> 
> ConfigExporter.java [warning] Line ~67-71: teamDef.getActionableItems() is called twice
> 
> Once during initial collection (line ~67) and again in the connections loop (line ~110). If getActionableItems() makes a server call or produces different results between invocations, this is a bug. If it's cached, it's just redundant work. Consider storing the result during the first pass.
> 
> [warning] Line ~73-76: getLeads and getMembers called twice per team def
> 
> Same issue as above — these are called during collection (lines 68-69) and again in the connections loops (lines 117-127). This doubles the API calls for every team definition.
> 
> [suggestion] Line ~85: addConfigEnumDefs only adds BuildImpactState
> 
> Per the design doc, enum defs should be generated from enumerated attribute types (e.g., ChangeType, WorkType, ClosureState). Currently only BuildImpactState is hard-coded. This appears to be an incomplete implementation for Phase 1 — consider adding a TODO comment.
> 
> [suggestion] Lines ~145-184: Part def construction is entirely hard-coded
> 
> The design doc states that attributes should be "derived from AtsArtifactTypes definitions" via type introspection. The current implementation manually lists attributes. This is pragmatic for Phase 1 but diverges from the stated approach. A comment acknowledging this would clarify intent.
> 
> WorkflowExporter.java [error] Line ~75-83: The query collects all team workflows but doesn't filter by state
> 
> For large programs with thousands of completed workflows, this query could return an enormous dataset without pagination or limits. The REST endpoint has no query parameters to control scope (e.g., only open workflows, date range). This is a potential performance and memory issue.
> 
> [warning] Line ~130-137: BID connections only check isProblemReport() workflows
> 
> The BuildImpactDataToTeamWf connection (BIT → CR direction) is never emitted. The code only outputs ProblemReportToBid (PR → BIT), but per the design doc, the BIT→CR connections (BuildImpactDataToTeamWf) should also be exported by iterating BID artifacts and their related CRs.
> 
> [warning] Line ~113-114: Action-to-workflow connection uses action.getId() as source name
> 
> The SysmlConnection source name is String.valueOf(action.getId()), but actions' part usages are also named by their ID (line ~176). This is consistent but uses numeric IDs as SysML part names, which are valid but not human-readable. The design doc examples use 'ACTION_12345' format with the ATS ID.
> 
> [suggestion] Line ~154: State machine name uses wd.getName() which may not be unique
> 
> If two team defs share the same WorkDefinition object, the Set correctly deduplicates. But if different WorkDefinition objects have the same name, both get added with duplicate state machine names in the output.
> 
> [suggestion] Lines ~97-106: Connection defs are added unconditionally
> 
> All connection defs (Derive, ResolvedBy, etc.) are emitted in the output even if no connections of that type exist in the data. While not incorrect, it adds noise. Consider only adding connection defs that have corresponding connection instances.
> 
> ExportProgramToSysmlNavigateItem.java [warning] Line ~35-36: Hard-coded output directory
> 
> System.getProperty("user.home") + "/SawPlProgramSysMl" is not configurable. There's a TODO comment acknowledging this. Per Java coding standards: "No hard-coded absolute file paths in production code."
> 
> [warning] Line ~49-50: outputDir.mkdirs() return value ignored
> 
> mkdirs() returns false if the directory couldn't be created. The code should check this and report failure.
> 
> [error] Line ~54: Hard-coded to DemoArtifactToken.SAW_PL_Program
> 
> This navigate item is registered in the general ATS navigation view (not a demo-only view), but it only works with the SAW PL demo program. Users without demo data will get a confusing error. Either gate registration behind a demo check, or provide a program selector dialog.
> 
> [suggestion] Line ~57: No error handling for Lib.writeStringToFile
> 
> If the write fails, the exception is caught by the outer catch block, but the error message will be generic. Consider catching IOException specifically for file operations.
> 
> AtsActionEndpointApi.java (new endpoints) [suggestion] Lines ~263-270: Missing @consumes annotation on GET endpoints
> 
> The getSysmlConfig and getSysmlWorkflows endpoints are GETs producing TEXT_PLAIN. This is fine for the response, but unusually, these take an ArtifactId path param. No issues here, but consider whether JSON output with APPLICATION_JSON would be more useful for programmatic consumers (text is fine for file download).
> 
> AtsActionEndpointImpl.java (new endpoints) [suggestion] Lines ~917-930: No error handling for export failures
> 
> If ConfigExporter.export() or WorkflowExporter.export() throws (e.g., program not found), the exception will propagate as an unhandled 500 error. Consider catching OseeArgumentException and returning a proper 404/400 response.

Done

dondunne:
Resolved David's Kiro comments and added some REST tests.

Grumb1e (approved):

david-w-miller (approved):

PR #656, opened 2026-09-03
Jaden/create artifact defaults and add another

jadenpuckett:


dondunne (dismissed):

Grumb1e (dismissed):

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/types/TypesEndpointImpl.java
```diff
@@ -113,9 +113,34 @@ public Collection<AttributePojo<?>> getArtifactTypeAttributes(ArtifactId artifac
       return orcsApi.tokenService().getArtifactTypes().stream().filter(
          art -> art.getId().equals(artifactId.getId())).flatMap(
             art -> art.getValidAttributeTypes().stream().map(
-               attr -> AttributePojo.valueOf(Id.SENTINEL, attr, GammaId.SENTINEL, "", "",
-                  art.getMultiplicity(attr)))).collect(
-                     Collectors.toList());
+               attr -> buildDefaultValuedAttribute(art, attr))).collect(
```
audrey-denk, line 116: I'm not sure, but I seem to remember Baldwin did something in MIM that is contrary to the rest of OSEE in that I believe he introduced some defaults as null.  I don't know if this will change that behavior, but just popped into my head and we may want to verify creating new elements/messages etc in MIM still work with this change.

audrey-denk (approved):
Left a comment in regards to the typesendpoint change.  May not be a biggie but worth looking into.

plugins/org.eclipse.osee.orcs.rest/src/org/eclipse/osee/orcs/rest/internal/types/TypesEndpointImpl.java
jadenpuckett: Good catch to flag it, but I don't think we need to change anything here — I verified it both by tracing the code and running MIM end-to-end.

On the code: the only thing this change adds is seeding the returned AttributePojo's value with the attribute's default via defaultValueAsString(...). That method just reads artType.getAttributeDefault(attrType) and returns "" when the default is null — it doesn't declare or change how defaults are defined. So for any attribute with a null default (including the MIM null-defaults Baldwin set up), the output is identical to the old code, which hardcoded "". Attributes with a real default (e.g. Extension = "md" on Markdown types) now come back pre-populated; null-default attributes are unchanged.

It's also not in the MIM creation path. This only affects GET /orcs/types/artifact/{id}/attributes (getArtifactTypeAttributes), whose consumers are the new-artifact-editor pre-population flows in the web app (artifact explorer create-child dialog, attributes-editor-panel "Add Attribute", actra workflow editor). MIM elements/messages are created through their own messaging services, not this endpoint.

To be safe I ran the full MIM ICD creation e2e chain against the branch build (backend + web) and all 12 steps pass — create connection, message + submessage, structure, elements, and array element. So MIM creation still works as expected with this change.

Grumb1e (approved):

PR #663, opened 2026-09-14
feature[TW29252]: Implement Markdown change report

jadenpuckett:
add paginated change report endpoint

update change report page to use paginated and filterable table

add branch picker to change report and md change report pages

move branch selector into icon button

fix filter and line numbering and branch page load bugs

add change report docs


audrey-denk (dismissed):

plugins/org.eclipse.osee.orcs.rest.model/src/org/eclipse/osee/orcs/rest/model/BranchEndpoint.java
```diff
@@ -222,12 +222,32 @@ int updateConflictStatus(@PathParam("branch") BranchId branch,
...
 
+   /**
+    * @deprecated Use {@link #getFilteredPaginatedChangeReport} for paginated, filtered changes.
+    */
+   @Deprecated
    @GET
    @Path("{branch1}/changes/{branch2}")
    @Produces({MediaType.APPLICATION_JSON})
    List<ChangeReportRowDto> getBranchChangeReport(@PathParam("branch1") BranchId branch1,
       @PathParam("branch2") BranchId branch2);
```
Grumb1e: Is there anything left that uses this? If so, deprecated works, otherwise I would prefer to remove it so it does not become lost and left around.

jadenpuckett: Good comment. I did not find anything using this endpoint, so I have removed it.

web/apps/osee/src/app/ple/markdown-diff/utils/export.ts
```diff
@@ -0,0 +1,272 @@
+/*********************************************************************
```
Grumb1e, line 1: I would suggest adding functions like 

```ts
function isNewEntry(entry: MarkdownDiffEntry): boolean {
	return entry.changeType === 'New' || !entry.wasValue.trim();
}

function isDeletedEntry(entry: MarkdownDiffEntry): boolean {
	return (
		entry.changeType.toLowerCase().includes('deleted') ||
		!entry.isValue.trim()
	);
}
```

To keep things consistent, readable, and "DRY".

jadenpuckett: done

Grumb1e (approved):

audrey-denk (approved):

PR #666, opened 2026-09-21
feature[TW29814]: Dynamically drive data rights

david-w-miller:


.kiro/steering/conditional/docs/dynamic-data-rights-classification.md
dondunne, line 1: Here's Kiro review.  I'd resolve them all, especially the blocking where production never calls refresh to load the db stored items.

---

I read the changed files plus the load-bearing production code that the diffs don't fully show: AttributeTypeEnum (before and after), DataRightsClassificationNameParser, DataRightsClassificationClientRefresher, and DataRightsClassificationAttributeType. Overall this is careful, well-tested work — the shared parser, the additive/idempotent refresh, and the ordinal-preserving promotion are correct and the concurrency reasoning holds. The design doc is genuinely useful.

There is one issue I'd call blocking, plus a few smaller ones.

Blocking
1. The client refresher appears to be dead code — nothing calls ensureRefreshed() in production.

DataRightsClassificationClientRefresher.ensureRefreshed() / reset() are only referenced by their own integration test. The design doc (and the class Javadoc) say the client promotes footer names "before building a DataRightsClassification pick-list" and resets "when client caches are cleared." But no production file in the PR wires either call:

The only modified skynet.core artifact is the new refresher class itself.
No pick-list construction path calls ensureRefreshed().
No client cache-clear path calls reset().
So on the client, footer-defined classifications will not show up in the attribute editor / mass editor pick-lists in normal use — the exact outcome the feature is for. The integration test passes only because it calls reset() + ensureRefreshed() directly, which masks the missing wiring.

Either the wiring got dropped from the branch, or it's intended for a follow-up. If it's a follow-up, the class and its test are premature and the doc overstates what ships. Please confirm and either add the call sites or scope the doc/PR to server-only. This is the one thing I'd block on.

Should fix
2. .gitignore now ignores all of .kiro/specs/ but the PR commits files under .kiro/steering/. That's fine as-is, but worth a sanity check that no intended spec artifacts were meant to ship. Minor, just calling it out.

3. AttributeTypeEnum.addDbLoadedValidEnum throws when enumTokens.isEmpty() ("no seed enum token available to clone"). For the data-rights type this can't happen (seeded from RequiredIndicator.values()), but the method is now public API on a general type. The guard is reasonable; I'd just confirm no caller can hit it with a legitimately empty enum type and get a surprising exception. Not a change request, just verify.

4. Broad catch (Exception) in two loaders. DataRightsFootersIndicatorLoaderImpl and the client refresher both catch Exception to cover "missing/unreadable/multiple-matches." The comments justify it and log at FINE, which matches the existing DataRightClassificationMap.create pattern, so I'm okay with it — but a catch this broad will also swallow programming errors (NPE, etc.) as "artifact missing." If the query layer exposes a narrower "not found" exception, prefer it. Otherwise leave the comment as-is since it explicitly documents the intent.

Nits (non-blocking)
JavaDoc volume vs. the standard. The house rule is "default to NO JavaDoc; add it only when it says something the code can't." Several new blocks are excellent (the addDbLoadedValidEnum promotion contract, the concurrency notes — keep those). But some restate the signature, e.g. DataRightsFooter.getClassification() / getContent() need none, and a few field comments on DataRightsOperationsImpl (cacheLock, requiredIndicatorRefresher) largely repeat the name. Trimming the redundant ones would bring it in line with the standard. Low priority.
Test performance. The property tests run 500–1000 iterations and the concurrency test dispatches 2000 tasks across 12 threads. They're seeded and deterministic (good), but that's real wall-clock cost on every suite run. Consider whether the iteration counts can drop (e.g. 100–200) without losing coverage. Judgment call.
AttributeTypeEnumTest @author Boeing — the standard says use a real first/last name. Minor.
Nice cleanups: removing the /* EOF */ trailers and switching size() == 0 to isEmpty() both match the standards doc.
What's solid (so you don't undo it)
The enumTokens → CopyOnWriteArrayList change plus synchronized on both mutators is the right, simple fix: writers are serialized so the size()-derived ordinal never races, and lock-free readers get stable snapshots. The concurrency test directly validates no duplicate names / no ordinal collisions.
Promotion moves the same token from dbLoadedEnumTokens to enumTokens rather than cloning — preserving ==/equals round-trip identity. That's the subtle correct choice and it's tested.
Extracting DataRightsClassificationNameParser as the single split rule (used by the map, the loader, and the client) removes the duplicated split("\n", 2) logic. Good deduplication.
Refresh triggers are limited to startup + cache-clear (not per-access), and DataRightsOperationsImplStartupTest proves the retrieval path never invokes the refresher. That's the performance-conscious design the task needs.
My recommendation: request changes, primarily to resolve item 1 (confirm/add the client wiring or rescope). The rest are cleanups that don't need to gate merge.

plugins/org.eclipse.osee.define.test/src/org/eclipse/osee/define/operations/publisher/datarights/DataRightClassificationMapTest.java
```diff
@@ -0,0 +1,195 @@
...
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Boeing - initial API and implementation
+ **********************************************************************/
+
+package org.eclipse.osee.define.operations.publisher.datarights;
+
```
dondunne, line 15: We're trying to get away from mockito tests, remove this whole test?  I think you just need to prove Data Rights loads once on client.

dondunne: I'd remove all these tests but one.

jadenpuckett (dismissed):

plugins/org.eclipse.osee.define.test/src/org/eclipse/osee/define/operations/publisher/datarights/DataRightClassificationMapTest.java
david-w-miller: I  checked these tests - they are JUnit only tests, the whole group runs in less than 3 seconds. I think they may capture and provide some context to AI (who is, in actuality, going to maintain them). So for this case I want to leave them in.

.kiro/steering/conditional/docs/dynamic-data-rights-classification.md
david-w-miller: I applied these comments, except the first. Since we don't change this data often, I accepted just needing an IDE restart to read the data. The refresh mechanism is implemented, and we could hook it up to any change of the well known data rights footer artifact.

dondunne (approved):

jadenpuckett (approved):

